PR #5, opened 2022-04-06
Include additional header files in wrapper.h

matsadler:
There are some functions part of Ruby's C API that are listed in their own headers rather than in `ruby.h`.

The one I really want is `ruby/encoding.h`, as I'm using it in Magnus, but when I took a look at it, it seemed like why not include the others.

Not all of these are guaranteed to exist, but the base `ruby.h` defines some macro constants when they exist, so I made use of that to conditionally include them.

### DIFF
--- a/crates/rb-sys/wrapper.h
+++ b/crates/rb-sys/wrapper.h
@@ -1,1 +1,70 @@
 #include "ruby.h"
+#ifdef HAVE_RUBY_ATOMIC_H
+#include "ruby/atomic.h"
+#endif
+#ifdef HAVE_RUBY_DEBUG_H
+#include "ruby/debug.h"
+#endif
+#ifdef HAVE_RUBY_DEFINES_H
+#include "ruby/defines.h"
+#endif
+#ifdef HAVE_RUBY_ENCODING_H
+#include "ruby/encoding.h"
+#endif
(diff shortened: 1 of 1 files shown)


ianks:
@matsadler can you rebase with `main` to trigger CI? i updated the triggers so it should run

matsadler:
It looks like the tests are filing on trying to run the tests for the example in `examples/simple`, but it looks like the same thing is happening in `main`.

I've taken a look, but I'm not really sure where the problem is yet.

ianks:
it looks unrelated to this PR, so I dont want to block from merging. Let me know if thats ok?

ianks:
I made some updates on `main` that should fix *some* of the issues.

matsadler:
I was seeing some errors in my Magnus `rb-sys` branch on Windows that looked like these changes were at fault, but I think it might be GitHub Action's `windows-latest` environment. The error didn't really make sense, plus went away when I switched to `windows-2019` 🤷

I am still having some errors with Windows, but I don't think it's these changes that are at fault.

I'm reasonably happy these changes are good, so if you're ok that the CI failures are something else could you merge this? (sorry, I know I have merge rights on this repo now, but I'm never sure on the etiquette merging my own PR)

ianks:
Looks good! I'll up CI in another branch

PR #7, opened 2022-04-10
Use current release of bindgen

matsadler:
As `allowlist_file` isn't in a released version of bindgen these changes do it the long way round with `allowlist_{function,type,var}`.

To get the list of functions/types/vars I built the `rb-sys` docs while using `allowlist_file`, copied out all the functions/etc names, then played around until I had regexes that would match them all, then transferred that to the `allowlist_{function,type,var}` calls. 

I diffed the resulting bindings from before and after the change, and they're pretty much identical, minus some constants prefixed `PRI` that I skipped on the assumption they were intended to be private.

I've also snuck in a couple of other commits, one to skip copying across the doc comments from the Ruby headers to the generated bindings. The Ruby doc comments don't translate to markdown so they're just a mess and not really useable.

I also changed the edition to 2018 as that, along with the change to bindgen 0.59, allows supporting back to Rust 1.46. This would allow things like using the system packaged Rust in the official Ruby docker image (based on Debian bullseye, which packages rust 1.48).

I can drop either/both these extra commits if they don't work for you.


matsadler:
I agree that `allowlist_file` would be easier to maintain, but I figured this would be a path to getting the crate in a releasable state without having to wait on bindgen releasing a new version with `allowlist_file` support, and it's easy to switch back once there is a bindgen release. 

While not as nice, I don't think this approach is a complete disaster for maintainability, I'd expect new stuff added to Ruby to be a bit more standardised than stuff added in the past, so be much more likely to fall under the regex patterns. Plus bindgen will recursively add types even if they don't match the allow list, so there's no danger of broken bindings where you allowlist a function but don't generate bindings for its arguments.

Here's a diff:
```diff
--- a/bindings_before.rs
+++ b/bindings.rs
@@ -171,7 +171,6 @@ pub const SIZEOF_DOUBLE: u32 = 8;
 pub const SIZEOF_TIME_T: u32 = 8;
 pub const SIZEOF_CLOCK_T: u32 = 8;
 pub const USE_UNALIGNED_MEMBER_ACCESS: u32 = 1;
-pub const PRI_LL_PREFIX: &[u8; 3usize] = b"ll\0";
 pub const HAVE_PID_T: u32 = 1;
 pub const SIGNEDNESS_OF_PID_T: i32 = -1;
 pub const HAVE_UID_T: u32 = 1;
@@ -186,9 +185,7 @@ pub const HAVE_MODE_T: u32 = 1;
 pub const SIGNEDNESS_OF_MODE_T: u32 = 1;
 pub const HAVE_RLIM_T: u32 = 1;
 pub const SIGNEDNESS_OF_RLIM_T: u32 = 1;
-pub const PRI_RLIM_PREFIX: &[u8; 3usize] = b"ll\0";
 pub const SIGNEDNESS_OF_OFF_T: i32 = -1;
-pub const PRI_OFFT_PREFIX: &[u8; 3usize] = b"ll\0";
 pub const HAVE_CLOCKID_T: u32 = 1;
 pub const SIGNEDNESS_OF_CLOCKID_T: u32 = 1;
 pub const HAVE_VA_ARGS_MACRO: u32 = 1;
@@ -203,8 +200,6 @@ pub const HAVE_DECL_GETENV: u32 = 1;
 pub const SIZEOF_SIZE_T: u32 = 8;
 pub const SIZEOF_PTRDIFF_T: u32 = 8;
 pub const SIZEOF_DEV_T: u32 = 4;
-pub const PRI_SIZE_PREFIX: &[u8; 2usize] = b"z\0";
-pub const PRI_PTRDIFF_PREFIX: &[u8; 2usize] = b"t\0";
 pub const HAVE_STRUCT_STAT_ST_BLKSIZE: u32 = 1;
 pub const HAVE_STRUCT_STAT_ST_BLOCKS: u32 = 1;
 pub const HAVE_STRUCT_STAT_ST_RDEV: u32 = 1;
@@ -246,10 +241,8 @@ pub const HAVE_INTPTR_T: u32 = 1;
 pub const SIZEOF_INTPTR_T: u32 = 8;
 pub const HAVE_UINTPTR_T: u32 = 1;
 pub const SIZEOF_UINTPTR_T: u32 = 8;
-pub const PRI_PTR_PREFIX: &[u8; 2usize] = b"l\0";
 pub const HAVE_SSIZE_T: u32 = 1;
 pub const SIZEOF_SSIZE_T: u32 = 8;
-pub const PRI_64_PREFIX: &[u8; 3usize] = b"ll\0";
 pub const HAVE_ALLOCA_H: u32 = 1;
 pub const HAVE_ALLOCA: u32 = 1;
 pub const HAVE_DUP: u32 = 1;
@@ -472,7 +465,6 @@ pub const CASEFOLD_FILESYSTEM: u32 = 0;
 pub const RUBY_MISSING_H: u32 = 1;
 pub const HAVE_FINITE: u32 = 1;
 pub const SIZEOF_VALUE: u32 = 8;
-pub const PRI_VALUE_PREFIX: &[u8; 2usize] = b"l\0";
 pub const RBIMPL_VALUE_NULL: u32 = 0;
 pub const RBIMPL_VALUE_ONE: u32 = 1;
 pub const RBIMPL_VALUE_FULL: i32 = -1;
@@ -550,28 +542,7 @@ pub const RB_SCAN_ARGS_LAST_HASH_KEYWORDS: u32 = 3;
 pub const RB_NO_KEYWORDS: u32 = 0;
 pub const RB_PASS_KEYWORDS: u32 = 1;
 pub const HAVE_RB_SCAN_ARGS_OPTIONAL_HASH: u32 = 1;
-pub const PRI_INT_PREFIX: &[u8; 1usize] = b"\0";
-pub const PRI_LONG_PREFIX: &[u8; 2usize] = b"l\0";
-pub const PRI_SHORT_PREFIX: &[u8; 2usize] = b"h\0";
 pub const RUBY_PRI_VALUE_MARK: &[u8; 2usize] = b"\x0B\0";
-pub const PRIdVALUE: &[u8; 3usize] = b"ld\0";
-pub const PRIoVALUE: &[u8; 3usize] = b"lo\0";
-pub const PRIuVALUE: &[u8; 3usize] = b"lu\0";
-pub const PRIxVALUE: &[u8; 3usize] = b"lx\0";
-pub const PRIXVALUE: &[u8; 3usize] = b"lX\0";
-pub const PRIsVALUE: &[u8; 4usize] = b"li\x0B\0";
-pub const PRIdPTRDIFF: &[u8; 3usize] = b"td\0";
-pub const PRIiPTRDIFF: &[u8; 3usize] = b"ti\0";
-pub const PRIoPTRDIFF: &[u8; 3usize] = b"to\0";
-pub const PRIuPTRDIFF: &[u8; 3usize] = b"tu\0";
-pub const PRIxPTRDIFF: &[u8; 3usize] = b"tx\0";
-pub const PRIXPTRDIFF: &[u8; 3usize] = b"tX\0";
-pub const PRIdSIZE: &[u8; 3usize] = b"zd\0";
-pub const PRIiSIZE: &[u8; 3usize] = b"zi\0";
-pub const PRIoSIZE: &[u8; 3usize] = b"zo\0";
-pub const PRIuSIZE: &[u8; 3usize] = b"zu\0";
-pub const PRIxSIZE: &[u8; 3usize] = b"zx\0";
-pub const PRIXSIZE: &[u8; 3usize] = b"zX\0";
 pub const USE_SYMBOL_AS_METHOD_NAME: u32 = 1;
 pub const RUBY_INTERN_H: u32 = 1;
 pub const INTEGER_PACK_MSWORD_FIRST: u32 = 1;
@@ -1309,6 +1280,12 @@ extern "C" {
 }
 #[repr(C)]
 #[derive(Copy, Clone)]
+pub union _bindgen_ty_1 {
+    pub x: *mut VALUE,
+    pub y: *const VALUE,
+}
+#[repr(C)]
+#[derive(Copy, Clone)]
 pub union _bindgen_ty_2 {
     pub x: *mut VALUE,
     pub y: *const VALUE,
@@ -1343,12 +1320,6 @@ pub union _bindgen_ty_7 {
     pub x: *mut VALUE,
     pub y: *const VALUE,
 }
-#[repr(C)]
-#[derive(Copy, Clone)]
-pub union _bindgen_ty_8 {
-    pub x: *mut VALUE,
-    pub y: *const VALUE,
-}
 extern "C" {
     pub fn rb_int2big(i: isize) -> VALUE;
 }
@@ -1635,10 +1606,10 @@ pub enum ruby_fl_type {
     RUBY_FL_USER18 = 1073741824,
     RUBY_FL_USER19 = -2147483648,
 }
-pub const RUBY_FL_DUPPED: _bindgen_ty_9 = _bindgen_ty_9::RUBY_FL_DUPPED;
+pub const RUBY_FL_DUPPED: _bindgen_ty_8 = _bindgen_ty_8::RUBY_FL_DUPPED;
 #[repr(u32)]
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
-pub enum _bindgen_ty_9 {
+pub enum _bindgen_ty_8 {
     RUBY_FL_DUPPED = 1055,
 }
 extern "C" {
```

I'm not sure why the `_bindgen_ty_`N types were shuffled around. I believe these are bindings to anonymous unions in C, but Rust doesn't allow anonymous types, so you end up with autogenerated names. While there might be cases where you'd want to reference their fields, I don't think you'd need to reference the type name.

ianks:
Ok this looks good @matsadler, can you clean up the conflicts and we can merge this??

crates/rb-sys/build.rs
```diff
@@ -128,13 +128,54 @@ fn main() {
...
+        .allowlist_var("_THREAD_SAFE")
+        .allowlist_var("__EXTENSIONS__")
+        .allowlist_var("__STDC_WANT_LIB_EXT1__")
         .rustified_enum("*")
         .blocklist_item("ruby_abi_version")
         .derive_eq(true)
         .derive_debug(true)
         .clang_args(clang_args)
         .parse_callbacks(Box::new(bindgen::CargoCallbacks))
+        .generate_comments(false)
```
ianks, line 178: Curious as to your thoughts on this, were comments causing an issue? Are they just not helpful?

matsadler: The comments for Ruby < 3.1 are pretty much worthless, but 3.1 added pretty nice documentation comments to the headers, they just don't translate across great.

I think when I first looked I happened to hit on one that was completely broken, but now I look again they are badly formatted, but I guess useable:
Screenshot 2022-04-18 at 21 11 50

I can drop this change if you prefer.

matsadler:
I've rebased on main.

I looked a little bit harder at the `PRI.*` values and I think the 'pri' is short for 'print', not 'private' as they seem to be used with `sprintf`, so I added them to the allow list too. With that change the only difference in the generated code is the shifting around of the `_bindgen_ty_`N types.

crates/rb-sys/build.rs
ianks: My vote would be to keep it for now, since it's at least somewhat readable on 3.1

ianks (approved):

PR #9, opened 2022-04-21
Use explicit usize for VALUE type

ianks:
# What?

This PR makes rb-sys use an explicity defined `VALUE` type instead of it being generated by bindgen.

# Why?

1. The logic of making `VALUE` a `usize` is fully compatible with a C ulong
2. `VALUE` is pervasive enough that it deserves it's own special type, with derivations and such
3. Using the same definition as ruby-core seems like a wise idea


matsadler (commented):
I haven't really looked at this before, and just accepted what bindgen spat out, but now I look I don't get why bindgen isn't using `usize` to begin with.

In Ruby's source [`VALUE` is defined as `uintptr_t`][def], which is apparently an integer type that can be round-tripped to a pointer and back. Rust's docs describe [`usize` as a pointer-sized integer][usize], which, oblique wording from the C standard aside, sounds like basically the same thing. I'm not really sure why bindgen is converting to `uintptr_t` to `u64` (and presumably `u32` on 32 bit platforms) and not `usize`.

[def]: https://github.com/ruby/ruby/blob/d1b9609cba8d7d59d6822e7fe16e311145e0db5e/include/ruby/internal/value.h#L40
[usize]: https://doc.rust-lang.org/stable/std/primitive.usize.html

So, switching to `usize` seems totally reasonable.

I'm not sure about defining `VALUE` as a newtype around `usize`.

On one hand I can see if you're writing code against rb-sys directly it's probably a little bit nicer to have an actual type where the compiler will error on you mixing `VALUE` and `usize`, rather than a type alias where they are just two names for the same thing.

But on the other hand, in C `VALUE` is just a typedef, so a type alias in Rust would hew closer the C api's line. If you're using rb-sys directly do you want something that most closely matches the C api?

It is kind of inconvenient for my use cases in Magnus, as I end up hitting a lot of the weird edges of Ruby's api, and with this change things like casting a pointer to `VALUE` would mean first casting it to a `usize`, and then wrapping it in a `VALUE`. Weirdly this would make my code *more* bound to `VALUE`'s underlying type as I have to know it contains a `usize`, rather that being able to pretend it's some opaque type I can cast to, when usually the new type pattern is trying to make code less aware of the contents.

But please don't put too much weight on my use case, it's not an insurmountable problem, it's not even a big one, it'd just be an annoying change for me to make when I switch over to rb-sys.

Maybe some methods for conversions like `from_ptr`, `as_ptr`, `from_usize`, and `as_usize` would be useful if you go the newtype route.

ianks:
I think it's probably not worth it if it's going to be more painful to integrate, so I'm going to close this for now. If we want to switch later, it's easy enough anyway.

Also, thanks for taking the time for the thoughtful response.

PR #20, opened 2022-04-29
Add rb-allocator

ianks:
This PR adds a wrapper around the [Rust system allocator](https://doc.rust-lang.org/std/alloc/struct.System.html) which reports memory usage back to the Ruby GC. This means that Ruby will have an accurate representation of Rust's memory usage, and thus know when to GC.

This crate exports, one type, `RbAllocator`, which implements the `GlobalAlloc` trait. It is a simple wrapper over the system allocator which reports memory usage to Ruby using `rb_gc_adjust_memory_usage`.

cc: @macournoyer @dylanahsmith


ianks:
IIRC, Windows does not have support for `atomic.h`, so not sure if `rb_gc_adjust_memory_usage` would work. Going to disable on Windows unless someone has a better idea!

crates/rb-allocator/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+//! Allocator which reports mem usage to the Ruby GC
+//!
+//! This crate exports, one type, `RbAllocator`, which implements the `GlobalAlloc` trait. It is a
+//! simple wrapper over the system allocator which reports memory usage to Ruby using
+//! `rb_gc_adjust_memory_usage`
+
+use rb_sys::{rb_gc_adjust_memory_usage, ssize_t};
+use std::alloc::{GlobalAlloc, Layout, System};
+
+pub struct RbAllocator;
```
matsadler, line 10: Could this wrap a generic allocator? Something like:

``` rust
pub struct RbAllocator<T>(T);

unsafe impl GlobalAlloc for RbAllocator<T>
where
    T: GlobalAlloc
{
    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
        let ret = self.0.alloc(layout);
```

crates/rb-allocator/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+unsafe impl GlobalAlloc for RbAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ret = System.alloc(layout);
+        if !ret.is_null() {
+            rb_gc_adjust_memory_usage(layout.size() as ssize_t)
```
matsadler, line 16: [The docs][gc.h] aren't super clear, and [the code][gc.c] is a bit beyond me but it _looks_ like this can trigger GC, such that it would make sense to call it _before_ the allocator, so that the CG can make room for the allocation.

How to account for the possibility of the allocation failing I'm not sure, `rb_gc_adjust_memory_usage` back down with a negative size I guess?

[gc.h]: https://github.com/ruby/ruby/blob/5c843a1a6e24aeabb3497065a362caf7b3e2d3b1/include/ruby/internal/intern/gc.h#L376-L388
[gc.c]: https://github.com/ruby/ruby/blob/5c843a1a6e24aeabb3497065a362caf7b3e2d3b1/gc.c#L11625

matsadler (commented):
This is a very cool idea!

It looks like Ruby actually [exports malloc, free, etc functions as part of its API][xmalloc.h], I wonder if you could take this idea to the next level and wrap Ruby's actual allocator.

[xmalloc.h]: https://github.com/ruby/ruby/blob/5c843a1a6e24aeabb3497065a362caf7b3e2d3b1/include/ruby/internal/xmalloc.h#L66-L86

crates/rb-allocator/src/lib.rs
dylanahsmith: `ruby_xmalloc` actually does do this actual the allocation, although I'm not sure why.  It does seem like it would make more sense for GC to be triggered before the allocation.

dylanahsmith:
> It looks like Ruby actually [exports malloc, free, etc functions as part of its API](https://github.com/ruby/ruby/blob/5c843a1a6e24aeabb3497065a362caf7b3e2d3b1/include/ruby/internal/xmalloc.h#L66-L86), I wonder if you could take this idea to the next level and wrap Ruby's actual allocator.

Ruby is really just wrapping the system allocator too.  That seems like the simpler solution, so perhaps that would result in some sort of incompatibility.  Was this not done to avoid raising a `NoMemoryError` when allocation fails?

crates/rb-allocator/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use rb_sys::{rb_gc_adjust_memory_usage, ssize_t};
+use std::alloc::{GlobalAlloc, Layout, System};
+
+pub struct RbAllocator;
+
+unsafe impl GlobalAlloc for RbAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ret = System.alloc(layout);
+        if !ret.is_null() {
```
dylanahsmith, line 15: If it does return null, then ruby's allocation functions (e.g. ruby_xmalloc) will invoke the garbage collector and retry.  I would have expected something like `rb_gc()` to be called to handle this condition.

```suggestion
        if ret.is_null() {
          rb_gc();
          ret = System.alloc(layout);
        }
        if !ret.is_null() {
```

along with an import of `rb_gc` and `ret` being made mutable.

examples/rust_reverse/test/test_rust_reverse.rb
```diff
@@ -10,4 +10,10 @@ def test_that_it_has_a_version_number
   def test_it_does_something_useful
     assert_equal "dlrow olleh", RustReverse.reverse("hello world")
   end
+
+  def test_stressing_it_out
+    10000.times do
+      assert_equal "a" * 100000, RustReverse.reverse("a" * 100000)
+    end
+  end
```
dylanahsmith, line 18: Another thing we lose from not using ruby's allocation functions is ability to use [`GC.stress=`](https://ruby-doc.org/core-3.1.2/GC.html#method-c-stress-3D) to force all the allocations to invoke the garbage collector for testing.  This can make it easier for ruby extension gems to be tested for properly marking values or using `RB_GC_GUARD(obj)` to make the code paths tested with GC stressed aren't going to access GC freed memory in production.

matsadler:
> Ruby is really just wrapping the system allocator too. That seems like the simpler solution, so perhaps that would result in some sort of incompatibility. Was this not done to avoid raising a `NoMemoryError` when allocation fails?

I was thinking of (but didn't actually put it in my last comment 🤷) cases where Ruby's been built with jemalloc, there's a Ruby configure script flag for this.

dylanahsmith:
A concern with invoking the ruby GC through the rust allocator is that it would add a lot of places where the GC could be invoked, which could result in issues from implicit aliasing that the rust compiler isn't aware of.

The GC being invoked means that the mark function for ruby data objects will be called, which could result in internal collections (e.g. Vec) being iterated.  Semantically, this is a read-only borrow reference, except it isn't made explicitly, so the rust borrow checker won't check for conflicts.  For instance, if a rust allocation is made while doing a mutable iteration over one of these collections, then doing a GC in this rust allocation hook would result in this usage being unsafe.

There is also the concern of making sure any ruby objects data being referenced is done with that ruby object being rooted.  E.g. if there is a reference to a string's bytes, then there needs to also be another reference to the ruby object that owns it that will be marked.

These reentrancy concerns are already concerns to be aware of when making calls to MRI C functions, but doing so for rust allocations makes it even more implicit.

An alternative would be to accumulate a memory usage delta in the `impl GlobalAlloc for RbAllocator` trait functions without calling `rb_gc_adjust_memory_usage` and expose a separate function on `RbAllocator` to manually call `rb_gc_adjust_memory_usage` with the accumulated delta if it isn't 0.  It would be nice to get this feature for free without users of the library having to think about it, but doing so by making it hard to write memory safe code seems counter to the reason why users would want to choose to write their ruby extension in rust.  It would still be easier to use in rust code expected to allocate/free memory compared to trying to calculate the delta and use `rb_gc_adjust_memory_usage` directly.

matsadler:
> The GC being invoked means that the mark function for ruby data objects will be called, which could result in internal collections (e.g. Vec) being iterated. Semantically, this is a read-only borrow reference, except it isn't made explicitly, so the rust borrow checker won't check for conflicts. For instance, if a rust allocation is made while doing a mutable iteration over one of these collections, then doing a GC in this rust allocation hook would result in this usage being unsafe.

I think if I'm understanding correctly, then this is an example of what you're talking about, written using Magnus:

``` rust
use std::cell::RefCell;

use magnus::{
    define_class, function, gc, method, DataTypeFunctions, Error, Module, Object, TypedData, Value,
};

#[derive(TypedData)]
#[magnus(class = "Vec", mark)]
struct RbVec(RefCell<Vec<Value>>);

impl DataTypeFunctions for RbVec {
    fn mark(&self) {
        // this will panic (in the middle of the garbage collector!) if
        // somewhere else has a `borrow_mut` still live.
        // `try_borrow` is of no use as we *must* mark the contents of the
        // vec, otherwise it could be GC'd
        gc::mark_slice(&self.0.borrow())
    }
}

impl RbVec {
    fn new() -> Self {
        Self(RefCell::new(Vec::new()))
    }

    fn push(&self, value: Value) {
        // This will allocate if the Vec needs to be resized, if allocation
        // triggers GC then `<RbVec as DataTypeFunctions>::mark` will be called
        // with the RefCell still mutably borrowed.
        self.0.borrow_mut().push(value)
    }
}

#[magnus::init]
fn init() -> Result<(), Error> {
    let class = define_class("Vec", Default::default())?;
    class.define_singleton_method("new", function!(RbVec::new, 0));
    class.define_method("push", method!(RbVec::push, 1));
    Ok(())
}
```

With Magnus once a Rust value is wrapped in a Ruby object Ruby will own that value. Magnus will only ever give you back a non-mutable reference, so you have to wrap the value in a type like `RefCef` to get mutability. In this case `RefCef` is doing the borrow checking at runtime, and could panic in the middle of GC. Panicking across FFI boundaries is undefined behaviour, so I'm going to adjust Magnus to catch the panic and abort for any case where code panics in the mark function.

Sorry @ianks for hijacking this thread with this diversion into Magnus stuff. Hopefully it's at least a useful illustration.
Thanks @dylanahsmith for outlining this edge case and helping me catch something I missed.

dylanahsmith:
Yes, that example is a useful more concrete example of the problem, which illustrates how easily the problem can occur.

ianks:
So if I'm understanding this correctly, maybe we can trigger [dont_gc_on](https://github.com/ruby/ruby/blob/3d6fd162a469f9bfa4afc696ef8b8d8536a38507/gc.c#L997) at the beginning of the allocation, so that ruby will not attempt to GC during the allocation. 

Obviously this isn't ideal, as it will not give ruby the chance to free up memory when rust allocates, but that's how things would work without this patch. However, ruby *would* have an accurate depiction of memory usage so it could free objects whenever an allocation is triggered outside of Rust.

Does this seam reasonable?

matsadler:
I don't think `dont_gc_on` is part of Ruby's public API. There's [`rb_gc_disable`], but if you [follow through the code][gc.c] that ends up finishing up the mark & sweep phases of the GC if they are being done incrementally, and that'll end up triggering exactly the thing you'd be trying to prevent.

[`rb_gc_disable`]: https://github.com/ruby/ruby/blob/master/include/ruby/internal/intern/gc.h#L288
[gc.c]: https://github.com/ruby/ruby/blob/84257244cd67e40daed274c01ff0a6929e1a2b67/gc.c#L11100

I experimented with this using the code above, and it does trigger the error, really easily.

---

I then tried with `rb_gc_adjust_memory_usage`, and I just can't get it to cause the problem.

Looking at the code for `rb_gc_adjust_memory_usage` again, `rb_gc_adjust_memory_usage` calls `objspace_malloc_increase` that calls `objspace_malloc_increase_body` which can start GC, but it only does so when the `type` argument is `MEMOP_TYPE_MALLOC`. `rb_gc_adjust_memory_usage` passes `MEMOP_TYPE_REALLOC` for `type`, so it won't trigger GC.

So, as far as I can tell, your original version with `rb_gc_adjust_memory_usage` is actually ok.

---

It might be worth also wrapping the `alloc_zeroed` and `realloc` functions of [GlobalAlloc](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html). By default the trait implements them in terms of `alloc` and `dealloc`, but it's possible allocators are providing their own optimised implementation. By not implementing them in your wrapper you'd be skipping those potential optimisations.

ianks:
> I then tried with `rb_gc_adjust_memory_usage`, and I just can't get it to cause the problem.
> 
> Looking at the code for `rb_gc_adjust_memory_usage` again, `rb_gc_adjust_memory_usage` calls `objspace_malloc_increase` that calls `objspace_malloc_increase_body` which can start GC, but it only does so when the `type` argument is `MEMOP_TYPE_MALLOC`. `rb_gc_adjust_memory_usage` passes `MEMOP_TYPE_REALLOC` for `type`, so it won't trigger GC.
> 
> So, as far as I can tell, your original version with `rb_gc_adjust_memory_usage` is actually ok.

After digging through the code myself, I had a similar hunch. I’ll try to confirm this is the expected behavior.

> It might be worth also wrapping the `alloc_zeroed` and `realloc` functions of [GlobalAlloc](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html). By default the trait implements them in terms of `alloc` and `dealloc`, but it's possible allocators are providing their own optimised implementation. By not implementing them in your wrapper you'd be skipping those potential optimisations.

Good call.

examples/rust_reverse/test/test_rust_reverse.rb
```diff
@@ -12,8 +12,12 @@ def test_it_does_something_useful
...
   def test_stressing_it_out
-    10000.times do
-      assert_equal "a" * 100000, RustReverse.reverse("a" * 100000)
+    GC.stress = true
+
+    expected = "a" * 10000
+
+    1000.times do
+      assert_equal expected, RustReverse.reverse("a" * 10000)
     end
```
dylanahsmith, line 21: You are missing `GC.stress = false` to avoid the `GC.stress = true` side-effect affecting all following tests.

dylanahsmith: Other than that, LGTM

PR #21, opened 2022-05-02
readme: Grammar fix "It's it not easy" -> "If it's not easy"

dylanahsmith:
I tried to preserve the original intention of what the sentence was trying to say.  Although, the emphasis on the word "easier" preceding that makes it seems like "easy" might not actually be the focus of this library.

Perhaps a more useful question to answer to understand the purpose of this library would be: When should rb-sys be used instead of alternatives (e.g. rutie or magnus)?


ianks:
🙏🏻 will adjust the readme based on your feedback as well!

PR #27, opened 2022-05-12
Add support for Ruby macros

ianks:
When working with Rust / Ruby, there are times when having these macros become incredibly useful. In particular, I found myself needing `RB_TYPE_P` which is a fairly optimized, non-trivial macro.

As such such, I think it makes sense for `rb-sys` to expose the low-level macros in Rust world, much as they would be used in C. Makes for a nice familiar experience when needed.

The implementation works by using the `cc` crate to compile a custom `libruby_macros.a` which is then linked, giving us access to the macros.


crates/rb-sys/Cargo.toml
```diff
@@ -25,3 +26,5 @@ doctest = false
 [build-dependencies]
 bindgen = "0.59"
 pkg-config = "0.3.24"
+cc = "1.0"
+shell-words = "1.1.0"
```
matsadler, line 30: Can these be optional and only activated when the `ruby-macros` feature is enabled?

Off the top of my head I think the syntax would be

```
[features]
ruby-macros = ["cc", "shell-words"]

[build-dependencies]
cc = { version = "1.0", optional = true }
shell-words = { version = "1.1.0", optional = true }
```

matsadler (approved):
Cool! Looks good to me 👍 

Might be nice to do this with some of the inline functions from Ruby's api too, `rb_array_len` for example.

crates/rb-sys/Cargo.toml
```diff
@@ -2,19 +2,20 @@
...
 edition = "2018"
-links = "ruby"
 readme = "readme.md"
 categories = ["ruby", "external-ffi-bindings"]
 description = "Rust bindings for the CRuby API"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/oxidize-rb/rb-sys"
 homepage = "https://github.com/oxidize-rb/rb-sys"
 documentation = "https://docs.rs/rb-sys"
+links = "rb" # we name it like this so users can still use magnus / rutie
```
matsadler, line 12: I've removed the `links = "ruby"` from Magnus and added a feature to use `rb-sys`, plus a couple of functions to make it easier to work with the two libraries, the changes are here: https://github.com/matsadler/magnus/commit/238487c3b08710b979ab37604e5c5e224af13b5a. I'm hoping to get a release out in the next week or so.

matsadler, line 12: I've removed the `links = "ruby"` from Magnus and added a feature to use `rb-sys`, plus a couple of functions to make it easier to work with the two libraries, the changes are here: https://github.com/matsadler/magnus/commit/238487c3b08710b979ab37604e5c5e224af13b5a. I'm hoping to get a release out in the next week or so.

PR #30, opened 2022-05-25
Remove dependency on libc crate

matsadler:
I might be missing some context, but it seems that as only basic types like `c_char`, `c_long`, etc are used, and we could instead use the ones from `std::os::raw` and skip the dependency.


ianks (approved):
Sounds good to me. Looks like they should be equivalent ([it is tested here](https://doc.rust-lang.org/1.40.0/src/std/os/raw/mod.rs.html#102-114)).

Issue #41, opened 2022-06-23
Add pipeline for testing with `ruby-static` feature in CI


ianks:
Currently, the Ruby versions published by ruby/setup-ruby [do not have libruby-static](https://github.com/ruby/setup-ruby/issues/142). So we'll need to compile our own version and test with the `--features ruby-static` flag.

ianks:
Done.

PR #43, opened 2022-06-23
Prevent libruby-static from linking unless `link-ruby` is specified

ianks:
So it looks like a `cargo:rustc-link-lib=static=ruby-static` was able to evade detection, causing rb-sys to link libruby-static even though the `link-ruby` flag was supplied. we don't actually want to link ruby unless we are in an embedded environment. Otherwise, we insta-segfault because the extension will be accessing the incorrect global symbols from libruby-static.


ianks:
resolves #41

PR #65, opened 2022-07-17
Fix in main readme file pointing to gem readme

dwdking:
URL is case sensitive, just corrected so it redirects properly


ianks:
thank you!

PR #72, opened 2022-07-25
[CI] Windows - test Windows MSYS2 builds with `stable-x86_64-pc-windows-gnu`, test mswin with `stable-x86_64-pc-windows-msvc`

MSP-Greg:
This is failing on the mswin build.  Not sure if there is a there is a missing dependency.


ianks:
Added a few commits hopefully to solve some issues with `atomic.h` on MSVC. Not sure why it was being included.

ianks:
Running into this now:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\lib\clang\13.0.1\include\emmintrin.h:2434:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
```

Need to [figure out a way to tell bindgen to use the windows include paths](https://github.com/rust-lang/rust-bindgen/issues/975)...

ianks:
@MSP-Greg do you know a reliable way to determine the sysroot of the msvc toolchain? once we are able to do that, we can tell clang to use it so it doesn't include the libclang headers. i'm a total windows newbie, so any help is greatly appreciated!

MSP-Greg:
@ianks

Years ago, I was writing C# code and occasionally using msvc.  Now, I mostly work with MSYS2 gcc or WSL2/Ubuntu gcc.  Or, I'm not that familiar with 'c' build systems, especially msvc.

So, I'm not sure.  But, enabling msvc sets all sorts of ENV variables, see:

https://github.com/oxidize-rb/rb-sys/runs/7560593302?check_suite_focus=true#step:4:4

For comparison, a list of ENV variables when msvc is not enabled is at:

https://github.com/MSP-Greg/actions-image-testing/runs/7561076503?check_suite_focus=true#step:3:6

JFYI, this all started when I noticed (in RubyGems) that targets `stable-x86_64-pc-windows-gnu` and `stable-x86_64-pc-windows-msvc` both existed. Given the naming, (gnu vs msvc), I suspected that using gnu with the Ruby mswin/msvc was probably a bad idea, or might be unstable.  The reverse was also true, using Ruby mingw/ucrt with msvc...

ianks:
It *seems* like this might be what we need:

```
INCLUDE: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.32.31326\ATLMFC\include;C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.32.31326\include;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um;C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\shared;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\um;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\winrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\cppwinrt
```

MSP-Greg:
Looking at the [log](https://github.com/oxidize-rb/rb-sys/runs/7560593302?check_suite_focus=true#step:4:22) from above, something seems odd.

The values of the `INCLUDE` and `LIB` ENV variables are similar to Windows `PATH`, but there are both single & double backslashes used for Ruby's equivalent of `SEPARATOR`.  This **might** be causing the problems?

ianks:
I added some logic to include all of the `INCLUDE` paths when generating the rust bindings on windows, hoping that fixes it 🤞🏻

MSP-Greg:
Thanks.  Re the issue above (backslash escaping?), I do have a fix for that. Not sure if it's a display issue or a value issue.

Anyway, since I'm rust challenged, I set `ENV['RUST_BACKTRACE']` to full, log is [here](https://github.com/MSP-Greg/rb-sys/runs/7621769809?check_suite_focus=true#step:4:121).  Not sure if that helps?  It doesn't include any of the changes you've added...

MSP-Greg:
Just looked at the run for `Debug emmintrin.h`.

This may be messy.  The file listed `/ucrt64/lib/gcc/x86_64-w64-mingw32/12.1.0/include/emmintrin.h` is part of the MSYS2 system, which probably shouldn't be used with msvc?  Not sure how it's finding it.  Somehow the build system is setting the paths?

ianks:
So I found `/ucrt64/lib/gcc/x86_64-w64-mingw32/12.1.0/include/emmintrin.h`, which 

> Thanks. Re the issue above (backslash escaping?), I do have a fix for that. Not sure if it's a display issue or a value issue.
> 
> Anyway, since I'm rust challenged, I set `ENV['RUST_BACKTRACE']` to full, log is [here](https://github.com/MSP-Greg/rb-sys/runs/7621769809?check_suite_focus=true#step:4:121). Not sure if that helps? It doesn't include any of the changes you've added...

I got past those issues. Seems like the current issue is finding a valid version of `emmintrin.h`, since right now it is defaulting to the clang version of the header which has an invalid definition for the `__m128i` type. I noticed one exists at `/ucrt64/lib/gcc/x86_64-w64-mingw32/12.1.0/include/emmintrin.h`, so just pushed a commit which tries to make bindgen use that when generating the Rust bindings for Ruby.

ianks:
> This may be messy. The file listed `/ucrt64/lib/gcc/x86_64-w64-mingw32/12.1.0/include/emmintrin.h` is part of the MSYS2 system, which probably shouldn't be used with msvc? Not sure how it's finding it. Somehow the build system is setting the paths?

Well, `bindgen` calls out to `libclang`, and it's actually finding ` C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\lib\clang\13.0.1\include\emmintrin.h`... Which results in this error:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\lib\clang\13.0.1\include\emmintrin.h:2374:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
```

Have no idea why as of now.

ianks:
Well, it's compiling now! But for some reason the tests are not executing....

ianks:
@MSP-Greg getting an error message about the makefile in CI:

```
fatal error U1036: syntax error : too many names to left of '='
```

any chance you could clone locally and debug?? or give some guidance about how to get a similar environment running in parallels?

MSP-Greg:
@ianks

We're working on releasing Puma 6.0.0, so that's taking a bit of my free time.  Let me see if I can work on it this weekend.

> give some guidance about how to get a similar environment running in parallels?

At a minimum, you'd need some version of Visual Studio, I think it's called 'Community Edition' (free) or something like that.  I've got VS2019 and VS2022 installed, usually work with 2022.  One enables the VS build tools via the command line, the provided script sets a lot of ENV variables.

Not sure about the other dependencies.  I'll hopefully see this weekend...

MSP-Greg:
@ianks

Locally, I can repo the error, which is explained simply as 'Only one string is allowed to the left of a macro definition'.
I haven't found a way to determine any more than that.

One thing that is a bit odd is that `bundle exec rake bindings:generate` is creating the file
`/tmp/bindings/rb-sys-0.9.29/3.2.0+2/x64-mswin64_140/bindings.rs`, which is equivalent to
`C:/tmp/bindings/rb-sys-0.9.29/3.2.0+2/x64-mswin64_140/bindings.rs`.

Not sure if that might be an issue, but `C:/tmp` isn't a 'standard' Windows folder.

ianks:
v0.9.31 now has support for mswin 🎊 ❤️

PR #80, opened 2022-09-28
Add suport for mswin

ianks:
#72 gets us closer to windows supports, insofar as not panicking immediately. I would like to get those changes in so I can continue to work on support.


ianks:
@MSP-Greg progress has been made. added support for `nmake`, now just trying to figure out how to wrangle `cl` into doing linking properly.

ianks:
The `Smoke test` will fail since it's using an older version of rb-sys. Going to merge since everything seems to be working otherwise 🎊

PR #81, opened 2022-09-29
Add a x86_64-linux-musl Dockerfile

eliias:
# Intent

To properly build for distributions like Alpine Linux, we need to be able to build a dynamic library for the x86_64-unknown-linux-musl target. The basic image is based on the rake-compiler-dock x86_64-linux image, but uses a Debian base instead of CentOS that provides the necessary `musl` dependencies https://github.com/messense/rust-musl-cross.

I have tried to separate the `rake-compiler-dock` related setup steps from the actual `rbsys` image, but we can also merge everything into this Dockerfile. The base image needs some further cleanup, but allow us to compile successfully today.

**Dockerfile for image**
https://github.com/y-crdt/yrb/blob/main/build/x86_64-linux-musl/x86_64-linux-musl.dockerfile

Opening primarily to start a discussion. @ianks looking forward to your thoughts.

## How to use

Your `Rakefile`:

```ruby
Rake::ExtensionTask.new("yrb", spec) do |ext|
  ext.source_pattern = "*.{rs,toml}"
  ext.cross_compile = true
  ext.cross_platform = %w[x86_64-linux x86_64-linux-musl]
end
```

Add the following to the `extconf.rb`:

```ruby
create_rust_makefile("yrb") do |r|
  if r.target&.include? "musl"
    r.extra_rustflags = %w[-C target-feature=-crt-static]
  end
end
```

Build gem:

```bash
rake gem:x86_64-linux-musl
```

Smoke test:

```bash
# I set the platform to properly emulate a x86_64 environment on a M1
docker run --platform linux/amd64 --name ruby-alpine -it ruby:alpine sh
docker cp pkg/y-rb-0.1.4.beta.1-x86_64-linux-musl.gem ruby-alpine:/y-rb.gem 
docker exec -it ruby-alpine bash sh
gem install rb_sys
gem install --local /y-rb.gem -v 0.1.4.beta.1
irb
require "y"
d1 = Y::Doc.new
t1 = d1.get_text("textfield_a")
t1 << "Hello, World!"
t1.to_s

d2 = Y::Doc.new
t2 = d2.get_tetx("textfield_a")
t2.to_s # should print ""
d2.sync(d1.diff)
t2.to_s # should print "Hello, World!"
```

Reference project: https://github.com/y-crdt/yrb
Published gems: https://rubygems.org/gems/y-rb/versions (`0.1.4.beta.1`)


eliias:
I need to fix the Rust setup one more time. I tried to convert over from my workaround to the `setup/rustup.sh` in this repo, but some linker setting aren't correct yet. With this image, the build links to `ld-linux-x86-64.so.2`, which isn't the goal.

docker/Dockerfile.x86_64-linux-musl
```diff
@@ -0,0 +1,30 @@
+FROM eliias/rake-compiler-dock-mri-x86_64-linux-musl:1.2.2
```
ianks, line 1: Is the plan to upstream this to `rake-compiler-dock`? In the meantime, can you copy-pasta the dockerfile content here?

ianks:
Awesome work! Excited to get this in 💪🏻. Couple of questions / comments:

```
 r.extra_rustflags = %w[-C target-feature=-crt-static]
```

Can this just be the default? If so, adding it the `CargoBuilder` as a default platform link arg should work.

- [ ] Can you add an entry to `data/toolchains.json` then run `rake data:derive`?
- [ ] Add an entry to the `readme.md`
- [ ] Once this is merged, we should add it to [oxi-test](https://github.com/oxidize-rb/oxi-test)

Thanks so much!

docker/Dockerfile.x86_64-linux-musl
eliias: No, not really. I just wanted to split between what is a somewhat 1:1 copy of the rake-compiler-dock unage and what is the actual rbsys part. I would need to copy a couple of files from rcd as well, do you think a subfolder for the Dockerfile is ok?

eliias:
> > > Can this just be the default? If so, adding it the `CargoBuilder` as a default platform link arg should work.
> 
> I just added this as an environment variable and I think it works.
> 
> ```
> CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS="-C target-feature=-crt-static"
> ```

This does not really work. I believe this to be a `cargo` issue and that it is not supported (I might be wrong, but I think what applies to compiler arguments might also be true for config files):
- # https://github.com/rust-lang/cargo/issues/10143
-  # https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/build_context/target_info.rs#L612

eliias:
@ianks I have moved all the Dockerfile dependencies and tested a build that was created with this image (it works). Addressed your other comments, but I assume you don't want to keep this image in sync with `rcd` forever. Maybe something we could address next?

docker/Dockerfile.x86_64-linux-musl
ianks: > and I believe the variable is reset in the Makefile):

I think setting `RB_SYS_EXTRA_RUSTFLAGS` should do the trick here

eliias:
@ianks Are you comfortable merging as is?

ianks (approved):

ianks:
Thanks again @eliias !

PR #83, opened 2022-10-10
Fix issues with cargo memory consumption in qemu environments

eliias:
On my M1 (using Docker for Mac) builds were intermittently failing due to containers running out of memory. It seems to always happen at the exact same step (fetching crates). Here is a similar issue documented in the cargo project https://github.com/rust-lang/cargo/issues/10583.


ianks:
🙏🏻

PR #95, opened 2022-11-07
Support global variable symbol resolution for `mswin`

ianks:
This PR addresses some issues on mswin with linking of global variables with the `__declspec(dllimport)` attribute. With this PR, mswin should be fully supported by rb-sys.


ianks:
cc: @MSP-Greg

PR #108, opened 2022-11-30
Include license file in rb-sys-build and rb-sys-env crates

ankane:
Hi, this includes license files in the published `rb-sys-build` and `rb-sys-env` crates. This copies the files (like the `rb-sys` crate):

```sh
cp LICENSE-* crates/rb-sys-build
cp LICENSE-* crates/rb-sys-env 
```

but could alternatively use symlinks.


ankane:
Thanks @ianks!

PR #109, opened 2022-11-30
Use proper bindgen include paths for LLVM on mswin

ianks:
Somewhere along the way, bindgen started including the wrong path for clang when generating bindings. This PR adds some debugging helpers, but ultimate relies on https://github.com/oxidize-rb/actions/commit/27945b582cd8f0e17392e8d0690012979923777e for the bindgen fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
         run: "bundle exec rake bindings:generate"
 
       - uses: actions/upload-artifact@v3
-        if: failure()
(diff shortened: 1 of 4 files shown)


ianks:
I'm noticing `LIBCLANG_PATH=C:\Program Files\LLVM\bin\`, but bindgen is including ` C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\lib\clang\15.0.1\include\emmintrin.h`...

PR #111, opened 2022-12-11
Block `compress-debug-sections` link arg

ianks:
Started seeing this pop up recently, [like here](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3661541694/jobs/6189849318).

I'm assuming this is happening because Ruby is adding this flag because it detects that `ld` supports it. However, in `rake-compiler-dock` we use a different `ld` for cross compilation which does not support the flag.

The benefits of it in Rust extensions are not clear, and it's better to let cargo choose anyway.

For reference, [there's a discussion on Ruby issue tracker about the flag](https://bugs.ruby-lang.org/issues/12934) and how it can negatively impact there performance of debugging tools, as well.

### DIFF
--- a/crates/rb-sys-build/src/rb_config.rs
+++ b/crates/rb-sys-build/src/rb_config.rs
@@ -26,6 +26,7 @@ pub struct RbConfig {
     pub link_args: Vec<String>,
     pub cflags: Vec<String>,
     pub blocklist_lib: Vec<String>,
+    pub blocklist_link_arg: Vec<String>,
     use_rpath: bool,
     value_map: HashMap<String, String>,
 }
@@ -41,6 +42,7 @@ impl RbConfig {
     pub fn new() -> RbConfig {
         RbConfig {
             blocklist_lib: vec![],
+            blocklist_link_arg: vec![],
             search_paths: Vec::new(),
             libs: Vec::new(),
             link_args: Vec::new(),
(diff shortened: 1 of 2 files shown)


jbourassa:
😞 looks like that didn't work for wasmtime-rb: https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3676458218/jobs/6217134014

ianks:
This update would be for the `rb-sys` _crate_. Can you try with the crate updated?

jbourassa:
> This update would be for the `rb-sys` _crate_. Can you try with the crate updated?

The commit for the above build should have the gem and crate on v0.9.48, which has this change -- right?

>```
> oxidize-rb/actions/setup-ruby-and-rust@main
> [...]
> Vendoring rb-sys v0.9.48 (/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rb-sys-0.9.48) to /home/runner/work/wasmtime-rb/wasmtime-rb/tmp/cargo-vendor/rb-sys
> Vendoring rb-sys-build v0.9.48 (/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rb-sys-build-0.9.48) to /home/runner/work/wasmtime-rb/wasmtime-rb/tmp/cargo-vendor/rb-sys-build
> Vendoring rb-sys-env v0.1.1 (/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rb-sys-env-0.1.1) to /home/runner/work/wasmtime-rb/wasmtime-rb/tmp/cargo-vendor/rb-sys-env
> [...]
>```

ianks:
Ahh. Weird. I’ll take another look.

PR #123, opened 2022-12-16
Block libssp

jbourassa:
wasmtime-rb builds are failing on Windows because libssp can't be found ([example build](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3707904651/jobs/6285025179)):
> error: could not find native static library `ssp`, perhaps an -L flag is missing?

Here's the result of [`RB_SYS_DEBUG_BUILD`](https://gist.github.com/jbourassa/4943fd60164a4c64236053b65b319bec). I found that libssp is on the machine, but not in the library path:
- `/usr/lib/gcc/x86_64-w64-mingw32/9.3-win32`
- `/usr/lib/gcc/x86_64-w64-mingw32/9.3-posix`

A [recent build on yrb](https://github.com/y-crdt/yrb/actions/runs/3696220120/jobs/6259625710) linked against `ssp` just fine. The 2 builds must use slightly different containers I assume (0.9.50 vs 0.9.51), not sure what that should change.

Skipping libssp fixed the issue, so I figured I should bring this up in a PR. **I don't know if it's the right fix.**

### DIFF
--- a/gem/lib/rb_sys/cargo_builder/link_flag_converter.rb
+++ b/gem/lib/rb_sys/cargo_builder/link_flag_converter.rb
@@ -7,7 +7,8 @@ module RbSys
     # Converts Ruby link flags into something cargo understands
     class LinkFlagConverter
       FILTERED_PATTERNS = [
-        /compress-debug-sections/ # Not supported by all linkers, and not required for Rust
+        /compress-debug-sections/, # Not supported by all linkers, and not required for Rust
+        /libssp/
       ]
 
       def self.convert(args)


jbourassa:
Looks like that did it for wasmtime-rb: https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3713714172

ianks:
Given ssp exists and is expected to be part of `LIBS`, I don't think the answer here is not ignore it.

You can see when invoking `make` on a normal C extension, it's an expected lib:

```
x86_64-w64-mingw32-gcc -shared -o testing.so testing.o -L. -L/usr/local/rake-compiler/ruby/x86_64-w64-mingw32/ruby-3.0.0/lib -static-libgcc -L. -pipe -s -pipe -s -Wl,--enable-auto-image-base,--enable-auto-import testing-x64-mingw32.def   -lx64-msvcrt-ruby300  -lshell32 -lws2_32 -liphlpapi -limagehlp -lshlwapi -l:libssp.a
```

`-l:libssp.a` should link. The question is, why can't it be found? The linker seems to be able to find it just fine:

```
x64-mingw32 ➜ 3.0.0 (main ✗) $ x86_64-w64-mingw32-gcc --print-file-name libssp.a
/usr/lib/gcc/x86_64-w64-mingw32/9.3-posix/libssp.a
```

So the question is, what can't find it -- Rust, some other tool? That's what we need to answer first....

ianks:
I believe `/usr/lib/gcc/x86_64-w64-mingw32/9.3-posix` is a special path that GCC knows about, but rust doesn't. Best solution in this case I believe to avoid transform libs in the form `/^-l\s*:lib(\S+).a$/` ([in this file](https://github.com/oxidize-rb/rb-sys/blob/621a3b44b12c38489409cb5eacf880fdb1a236e5/gem/lib/rb_sys/cargo_builder/link_flag_converter.rb#L27)). I.e. what used to be `-l static=ssp` should just be `-C link-arg=-l:libssp.a`.

Wann give that a whirl?

jbourassa:
Thanks for the hint!

> Wanna give that a whirl @jbourassa ?

Absolutely, let me try that 👍.

jbourassa:
Closing in favour of #124

PR #124, opened 2022-12-17
Stop rewriting static libs

jbourassa:
As per Ian's [suggestion here](https://github.com/oxidize-rb/rb-sys/pull/123#issuecomment-1355635048), this PR makes it so `-l:lib.+\.a$` are not rewritten, and instead passed as link-arg.

That seems to work in wasmtime-rb as per my `rb-sys-dock` experiments. I haven't tried the full rb-sys in CI yet with the crate changes.

### DIFF
--- a/crates/rb-sys-build/src/rb_config.rs
+++ b/crates/rb-sys-build/src/rb_config.rs
@@ -275,7 +275,6 @@ impl RbConfig {
         let search_path_regex = Regex::new(r"^-L\s*(?P<name>.*)$").unwrap();
         let lib_regex_short = Regex::new(r"^-l\s*(?P<name>\w+\S+)$").unwrap();
         let lib_regex_long = Regex::new(r"^--library=(?P<name>\w+\S+)$").unwrap();
-        let static_lib_regex = Regex::new(r"^-l\s*:lib(?P<name>\S+).a$").unwrap();
         let dynamic_lib_regex = Regex::new(r"^-l\s*:lib(?P<name>\S+).(so|dylib|dll)$").unwrap();
         let framework_regex_short = Regex::new(r"^-F\s*(?P<name>.*)$").unwrap();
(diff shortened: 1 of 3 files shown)


crates/rb-sys-build/src/rb_config.rs
```diff
@@ -495,30 +492,6 @@ mod tests {
...
-    }
-
-    #[test]
-    fn test_unconventional_lib_with_colon() {
-        let mut rb_config = RbConfig::new();
-        rb_config.push_dldflags("-l:ssp.a");
-
-        assert_eq!(rb_config.link_args, vec!["-l:ssp.a"]);
-    }
-
```
jbourassa, line 521: These felt irrelevant now so I removed them.

gem/lib/rb_sys/cargo_builder/link_flag_converter.rb
```diff
@@ -22,14 +22,12 @@ def self.convert_arg(arg)
...
           ["-l", $1]
-        when /^-l\s*:lib(\S+).a$/
-          ["-l", "static=#{$1}"]
         when /^-l\s*:lib(\S+).(so|dylib|dll)$/
           ["-l", "dylib=#{$1}"]
         when /^-F\s*(.*)$/
           ["-l", "framework=#{$1}"]
         else
-          ["-C", "link_arg=#{arg}"]
+          ["-C", "link-arg=#{arg}"]
```
jbourassa, line 30: Couldn't find any reference to `link_arg` (underscored), I _think_ this should be `link-arg`.

ianks: oh my nice find 😮

ianks (approved):

jbourassa:
Thanks for publishing a new version. wasmtime-rb's build gem step [now passes](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3720542004) 🎉 !

PR #125, opened 2022-12-24
Fix build on Ruby 3.2

peterzhu2118:
After ruby/ruby@cd1a0b3, only development versions of Ruby have RUBY_ABI_VERSION defined. This means that rb-sys fails to compile on released version of 3.2.0.


ianks:
Failures look unrelated to the patch, I’ll dive into them later. Ty!

byroot:
Sorry to be a bother on Xmas day, but are you planning to release this soon? It's preventing to upgrade to 3.2, and rb_sys can't be used as a git gem, so there's no workaround I can thing of.

ianks:
Yeah I’ll release this now

ianks:
Done, thanks for your patience!

PR #128, opened 2023-01-08
Add support for Ruby 3.2

ianks:
This PR makes some changes to support Ruby 3.2. I made the docker images inherit from the RCD snapshot image. My thinking is this will give us frequent updates, and stability should be OK.  @flavorjones do you have any thoughts on this?

Testing this build here: https://github.com/oxidize-rb/rb-sys/actions/runs/3864719313

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,8 +32,8 @@ jobs:
       # Test the container
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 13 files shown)


flavorjones:
Inheriting from the snapshot images seems like the right thing to do today. Once @larskanis cuts a rake-compiler-dock release  with 3.2 support, though, it might be better to switch to that one for stability.

PR #132, opened 2023-01-12
Use new RCD snapshot tag

ianks:
see https://github.com/rake-compiler/rake-compiler-dock/issues/96


ianks:
[⏳ Testing here...](https://github.com/oxidize-rb/rb-sys/actions/runs/3898807995)

PR #134, opened 2023-01-19
Support openbsd

marvinthepa:
Hey,

I know that OpenBSD is not one of your supported platforms, but I'll try anyway. I hope you are more responsive than the rutie maintainer :-).

Maybe it is okay to use `-fdecspec` for all platforms, But I don't know how to find out..

`-fmsextensions` fails on OpenBSD:

```
/usr/include/machine/_types.h:133:15: error: cannot combine with previous 'int' declaration specifier
```

https://zewaren.net/freebsd-10-build.html gave me the idea that this might have to do with `msextensions`.

### DIFF
--- a/crates/rb-sys-build/src/bindings.rs
+++ b/crates/rb-sys-build/src/bindings.rs
@@ -12,10 +12,16 @@ use std::process::Command;
 pub fn generate(rbconfig: &RbConfig, static_ruby: bool) {
     ensure_rustfmt_available();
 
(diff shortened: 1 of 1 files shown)


ianks:
Would love to support openbsd if possible, thanks for you contribution!

ianks:
The plan  is to port rutie to rb-sys. So FWIW, this patch would improve the openbsd story on rutie as well (when either I or someone else finds time)

PR #160, opened 2023-02-25
Hoist `list_platforms` above option parsing

gjtorikian:
I noticed the `--list` CLI command was erroring with a `NoMethodError`; it's because the method necessary to list the platforms is defined after the option parsing.


ianks:
Thank you!

PR #161, opened 2023-03-05
Allow to add a space between Ruby versions in `rb-sys-dock`

y-yagi:
I tried to specify `--ruby-versions` of `rb-sys-dock` like `'3.1, 3.2'` at the first time. But the current parse process treats space as a value too. So Ruby versions will be `["3.1.0", "..0", "3.2.0"]` at last. Of course, these are not expected values.

In my personal opinion, adding space between values is happened often. So this patch fixes to allow it.


ianks (approved):
Thank you!

PR #163, opened 2023-03-12
Keep other metadata when sets `rubygems_mfa_required`

y-yagi:
Currently, metadata is overridden when set `rubygems_mfa_required`.  So other metadata, like `source_code_uri` is lost. This fixes to keep all metadata.


ianks (approved):

ianks:
Great catch, thank you!

PR #167, opened 2023-03-20
Make `RbSys::ExtensionTask` work without gem spec

y-yagi:
Currently, if users use `RbSys::ExtensionTask` without a gem spec, `RbSys::ExtensionTask` raises `ArgumentError`.

This is because `Rake::BaseExtensionTask#initialize` calls `RbSys::ExtensionTask#init` with `gem_spec=nil`.
https://github.com/rake-compiler/rake-compiler/blob/f845901f79681ddee8d387c20212585dee494990/lib/rake/baseextensiontask.rb#L28-L29

So `gem_spec` it to be `nil`, not ` :undefined`. That's why `RbSys::ExtensionTask#lint_gem_spec` raises `ArgumentError`.

This fixed to override initialize and set `gem_spec` as expected.


ianks (approved):

ianks:
Thank you!

PR #171, opened 2023-03-22
Bump syn build dependency to 2.0

dylanahsmith:
Closes #168, which is a dependabot PR that includes commits and release notes for context


dylanahsmith:
CI is failing for ubuntu-latest, since syn 2.0 uses rust edition 2021, which would be unstable for an old version of rust packaged for linux distros.  So there may be a need to hold off on bumping syn while compatibility with those old rust versions is needed.

ianks:
Hmmm, our minimum supported rust version right now is 1.51. does that support 2021?

Issue #176, opened 2023-03-24
Drop support for Ruby < 2.7


ianks:
Currently, we support some old old rubies. As part of 1.0, we no longer will. So official support will be dropped.

- [ ] Remove entries from CI matrix
- [ ] Update docs
- [ ] Check Rubygems CI
- [ ] Remove any conditional code for old versions
- [ ] Update `SUPPORTED_RUBY_VERSIONS` const in the build/main.rs

PR #187, opened 2023-04-01
Drop support for Ruby < 2.7

jbourassa:
Addresses #176:

- [x] Remove entries from CI matrix
- [x] Update docs
  💭 I think I got all the docs, at least according to my `rg`-fu in this repo. Anything else?
- [x] Check Rubygems CI
  I ran [Ruby's test suite](https://github.com/jbourassa/ruby/actions?query=branch%3Arb-sys-drop-old-rubies) with `rb-sys` [pointing to this branch](https://github.com/jbourassa/ruby/commit/e835f5f780930f5aee31c0d9fd6990fb55acfc88), I think that works? Looks like CI failed for other reasons, the tests passed.
- [x] Remove any conditional code for old versions
- [x] Update `SUPPORTED_RUBY_VERSIONS` const in the build/main.rs

Let me know if you think anything is missing.


.standard.yml
```diff
@@ -1,7 +1,7 @@
 # For available configuration options, see:
 #   https://github.com/testdouble/standard
 ---
-ruby_version: 2.3
+ruby_version: 2.6
```
gjtorikian, line 4: If Ruby < 2.7 is not supported, should this be something higher?

phoenixflutterby (commented):
Weed The ScriptS.

PR #225, opened 2023-06-13
install extra rustup targets from makefile

ggmichaelgo:
### What are you trying to solve?

Add a new config for mkmf to install rustup target.
The gems that are targeting `wasm32-unknown-unknown` can install the required target in the makefile.


ianks:
❤️

PR #231, opened 2023-06-20
Update syn to 2.0

matsadler:
This takes the changes from #171 (Thanks @dylanahsmith, hope you don't mind), rebases them on main, and adds a couple more.

The motivation behind this is, I've updated Magnus to syn 2.0 and it'd be nice to have rb-sys and Magnus be able to use the same syn so it only need to be compiled once.

The biggest change is probably upping the minimum supported Rust version to 1.60, so that bindgen can be updated to 0.66, so that its syn dependency is upped to 2.0. Without this change the whole motivation for making any change is lost. Increasing the MSRV isn't something that bothers me right now as I'm going to be increasing it to 1.61 in Magnus, but if it doesn't work with your plans, then no worries, this whole change is just a nice to have, not critical, I'd be ok with dropping it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,7 +28,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -242,7 +242,7 @@ dependencies = [
  "quote",
  "regex",
  "shell-words",
- "syn",
+ "syn 2.0.5",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


dylanahsmith:
> This takes the changes from #171 (Thanks @dylanahsmith, hope you don't mind)

Much appreciated, since I wasn't planning on continuing that work after being laid off from my job at Shopify

ianks:
I'm 👍🏻  with this, 1.60 is available on latest ubuntu for users that don't use rustup.

PR #245, opened 2023-07-28
Remove all `Cargo.toml` files from spec.extensions

Youssef-Mak:
All `*/Cargo.(toml|lock)` files should be removed from `spec.extensions`.

### DIFF
--- a/gem/lib/rb_sys/extensiontask.rb
+++ b/gem/lib/rb_sys/extensiontask.rb
@@ -45,6 +45,7 @@ module RbSys
         gemspec.files.reject! { |f| f.end_with?(".rs") }
         gemspec.files.reject! { |f| f.match?(/Cargo.(toml|lock)$/) }
         gemspec.files.reject! { |f| extconf.end_with?(f) }
+        gemspec.extensions.reject! { |f| f.end_with?("Cargo.toml") }
       end
     end
 


gem/lib/rb_sys/extensiontask.rb
```diff
@@ -45,6 +45,7 @@ def init(name = nil, gem_spec = nil)
         gemspec.files.reject! { |f| f.end_with?(".rs") }
         gemspec.files.reject! { |f| f.match?(/Cargo.(toml|lock)$/) }
```
Youssef-Mak, line 46: Wouldn't it be already removed here? 🤔

gem/lib/rb_sys/extensiontask.rb
```diff
@@ -45,6 +45,7 @@ def init(name = nil, gem_spec = nil)
...
         gemspec.files.reject! { |f| extconf.end_with?(f) }
+        gemspec.files.reject! { |f| extconf.match?(/Cargo.(toml|lock)$/) }
```
Youssef-Mak, line 48: pretty sure `extconf.match?(/Cargo.(toml|lock)$/)` will never be true 😕

ianks: Thanks for your contribution! I think what we want is:

```suggestion
        gemspec.extensions.reject! { |f| f.end_with?("Cargo.toml") }
```

This will prevent rubygems from attempting to compile the gem with `cargo`, which is unnecessary for precompiled gems. Does that make sense?

Youssef-Mak: Didn't consider `gemspec.extensions` 😲  Yeah makes sense 👍

gem/lib/rb_sys/extensiontask.rb
```diff
@@ -45,6 +45,7 @@ def init(name = nil, gem_spec = nil)
         gemspec.files.reject! { |f| f.end_with?(".rs") }
         gemspec.files.reject! { |f| f.match?(/Cargo.(toml|lock)$/) }
         gemspec.files.reject! { |f| extconf.end_with?(f) }
+        gemspec.extensions.reject! { |f| f.end_with?("Cargo.toml") }
```
Youssef-Mak, line 48: Should we also consider `Cargo.lock` here?

ianks: I think this is good, since rubygems will deny any Cargo.lock anyway.

ianks:
🙇🏻

PR #247, opened 2023-08-01
Make docs more accessible

Youssef-Mak:
We want to make the docs a bit more welcoming. 

The main change here is to add a "New to `rb-sys`?" section that serves as an entry point for new visitors to have a quick view of the documentation that's available.

Not sure whether to put this section above `Features` 🤔


ianks:
Thank you so much!

PR #294, opened 2023-12-22
Jruby support

uvlad7:
Use forked oxi-test


uvlad7:
It appears that `actions/setup-java` breaks `oxidize-rb/actions/setup-ruby-and-rust` cache on macos - cargo suddenly becomes linux binary. I had to run `gh cache delete --all` to [bypass](https://github.com/uvlad7/rb-sys/actions/runs/7306198300/job/19910731044) this.

PR #307, opened 2024-01-03
Add musl support for Ruby 3.3

ianks:


ianks:
[Things seem to be working](https://github.com/oxidize-rb/rb-sys/actions/runs/7402601468/job/20140800278)

PR #321, opened 2024-01-23
RbConfig: check if config dump succeeded

ngroman:
Adds a defensive check to make sure the call to dump the `RbConfig::CONFIG` exited successfully.

This makes it easier to diagnose a failure due to bazel misconfiguration. Without this the failure manifests as an empty config map and later as `panicked at 'Key not found: MAJOR'`


ngroman:
Happy to riff on the error message wording here, but this would have definitely saved me a huge headache earlier today.

ianks (approved):

ianks:
Thanks for this, will ship it! ❤️

ianks:
`ruby-head` is busted for CI, not due to your changes]

PR #339, opened 2024-03-04
Use rake-compiler-dock musl images

ianks:
Goal here is to use `rake-compiler-dock` images for `*-musl`. Currently, they are in the 1.5.0rc. cc: @flavorjones @eliias

### DIFF
--- a/docker/Dockerfile.x86_64-linux-musl
+++ b/docker/Dockerfile.x86_64-linux-musl
@@ -1,150 +1,4 @@
-# provides musl cross-compile toolchain
-FROM messense/rust-musl-cross:x86_64-musl
-
-##########################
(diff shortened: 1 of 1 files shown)


ianks:
Images built properly and tests pass, can test now with `docker pull rbsys/x86_64-linux-musl:sha-7d345bd11d5ebf8f992118dcc3b0208d4d59eaed`

flavorjones:
1.5.0.rc1 should be fine from a functional perspective, the only thing blocking a final release is more/better documentation.

Speaking of which, please make sure to review https://github.com/rake-compiler/rake-compiler-dock?tab=readme-ov-file#gnu-and-musl for caveats.

PR #354, opened 2024-04-11
Mount current directory, then CD to the working directory

marstaik:
Currently the docker image mounts the provided working directory. However, when building extensions that require parent hierarchies, this doesn't work.

Consider this example

```
api/ruby/
  Gemfile
  Cargo.toml
packages/core/
  Cargo.toml
Cargo.toml
```

The ruby gem's `Cargo.toml` requires the relative `../../packages/core/Cargo.toml` to complete building.
For the bundler to execute, you would pass the -directory (or from the actions working_directory argument) as `api/ruby`. However, this currently only results in the working_directory folder being mounted, making it so the build cannot complete.

I believe the most logical approach is to mount the current directory, and then change directories.

### DIFF
--- a/gem/exe/rb-sys-dock
+++ b/gem/exe/rb-sys-dock
@@ -359,7 +359,7 @@ def rcd(input_args)
   cmd = <<~SH
     #{default_docker_command} run \
       --platform #{OPTIONS.fetch(:docker_platform)} \
-      -v #{working_directory}:#{working_directory} \
+      -v $(pwd):$(pwd) \
       #{mount_tmp_dir} \
       #{mount_target_dir} \
       #{mount_cargo_registry} \


mkdynamic:
Bisecting versions, this broke building for me. I'm not sure why.

With no other changes, running `bundle exec rake "native[arm64-darwin]"` on `0.9.91` works. And fails on `0.9.92` with the following error:

```bash
$ bundle exec rake "native[arm64-darwin]"

🐳 Running default build command (rake native:arm64-darwin gem)
🔍 Running command:
	$ /users/mark/.docker/bin/docker run --platform linux/amd64 -v $(pwd):$(pwd) ...

rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)

(See full trace by running task with --trace)
rake aborted!
Command failed with status (1): [bundle exec rb-sys-dock --platform arm64-darwin --build --verbose]
/users/mark/monorepo/libs/mail_auth/Rakefile:16:in block in <top (required)>'

...
```

My project is a monorepo, with a layout like this:

```
monorepo/
  .git/
  libs/
    mail_auth/
      Rakefile
      ...
```

On MacOS arm64.

Any clues?

>mount the current directory, and then change directories

Where does it actually `cd`?

PR #394, opened 2024-07-01
Allow missing AR/CC/etc tools for precompiled rubies

ianks:
Sometimes the AR setup in `rbconfig.rb` can incorrect in the case of precompiled binaries. This PR makes us resistant to that.


crates/rb-sys-build/src/cc.rs
```diff
@@ -270,7 +273,13 @@ fn get_tool(env_var: &str, default: &str) -> Command {
     let mut tool_args = shellsplit(tool_args).into_iter();
     let tool = tool_args.next().unwrap_or_else(|| default.to_string());
 
-    let mut cmd = new_command(&tool);
+    let mut cmd = if !Path::new(&tool).is_file() {
```
ianks, line 276: This is the fix

driv3r: imho I prefer to avoid negation in conditionals if possible, so instead I would go with 
```rust
let mut cmd = if Path::new(&tool).is_file() { 
    new_command(&tool)
} else {
    debug_log!("[WARN] {tool} tool not found, falling back to {default}");
    new_command(default)
}
```

feels easier to read/undertsand, but just an opinion xD

driv3r (approved):
Looks legit

crates/rb-sys-build/src/cc.rs
ianks: done!

ianks:
The CI failures seems GHA related, so I'm going to merge with them failing for now.

PR #404, opened 2024-07-26
Silence new unexpected_cfgs lint

matsadler:
Rust 1.80 got a new [lint to warn on unexpected cfg](https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#checked-cfg-names-and-values), which generates a lot of noise building both Magnus and rb-sys.

To make cfg expected we either need to output `cargo::rustc-check-cfg=cfg(name)` from build.rs or add:

``` toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(name)'] }
```

to Cargo.toml.

Both `cargo::rustc-check-cfg` and `cargo:rustc-check-cfg` (two or one colons) are accepted. `::` is the 'correct' syntax as of Rust 1.77, but will error on some earlier versions. The `:` syntax is accepted (with warnings on some older versions) with both old and new versions, so these changes use `:`.

### DIFF
--- a/crates/rb-sys-build/src/bindings.rs
+++ b/crates/rb-sys-build/src/bindings.rs
@@ -192,6 +192,10 @@ fn push_cargo_cfg_from_bindings(
             if is_defines(&conf_name) {
                 let name = conf_name.to_lowercase();
                 let val = conf.value_bool().to_string();
+                println!(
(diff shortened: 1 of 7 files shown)


ianks:
Thank you, Mat! The failures seem to be unrelated so I'll merge and cut a release next week

PR #408, opened 2024-08-01
fix: openssl can't build on rb-sys-dock

### LINKED ISSUE #402: rb-sys-dock's base image is past EOL
rb-sys-dock depends on CentOS 7, which as of Jul 1 is EOL.

For me, this manifested as an inability to build OpenSSL dependencies: I need to `yum install perl-IPC-Cmd` to make `x86_64-linux` builds work, and that fails with this error:

```
Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
```

gh workflow logs: https://github.com/BoundaryML/baml/actions/runs/10066707009/job/27828665640

centos mirrorlist deprecation: https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve

sxlijin:
Fixes #402 , I think - I don't know how to test this without merging this PR and pushing out the release.

### DIFF
--- a/docker/Dockerfile.aarch64-linux
+++ b/docker/Dockerfile.aarch64-linux
@@ -1,4 +1,4 @@
-FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.5.1-mri-aarch64-linux
+FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.5.2-mri-aarch64-linux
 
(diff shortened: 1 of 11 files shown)


ianks (approved):

ianks:
Thank you! will push this out today

PR #410, opened 2024-08-02
Try to remove llvm-15 patch

ianks:


stanhu:
@ianks It looks like we can close this with https://github.com/oxidize-rb/actions/pull/55. 😄

PR #430, opened 2024-10-14
Use only ARGV after `--` in `create_rust_makefile`

y-yagi:
I sometimes use Rake with `--trace` option, but it fails with when loading `create_rust_makefile`.

```bash
$ bundle exec rake --trace
** Invoke default (first_time)
** Invoke test (first_time)
** Invoke compile (first_time)
** Invoke compile:x86_64-linux (first_time)
(omit)
cargo rustc  --manifest-path /home/y-yagi/src/github.com/y-yagi/watchcat/ext/watchcat/Cargo.toml --target-dir /home/y-yagi/src/github.com/y-yagi/watchcat/target --lib --profile release --trace -- -C linker=gcc -L native=/home/y-yagi/.rbenv/versions/3.3.4/lib -L native=/home/linuxbrew/.linuxbrew/opt/gmp/lib -C link-arg=-Wl,-rpath,/home/linuxbrew/.linuxbrew/opt/gmp/lib -C link-arg=-lm -l pthread
error: unexpected argument '--trace' found

  tip: a similar argument exists: '--crate-type'
  tip: to pass '--trace' as a value, use '-- --trace'

Usage: cargo rustc --manifest-path <PATH> --target-dir <DIRECTORY> --lib --profile <PROFILE-NAME> --crate-type <CRATE-TYPE> [ARGS]...

For more information, try '--help'.
make: *** [Makefile:566: /home/y-yagi/src/github.com/y-yagi/watchcat/target/release/libwatchcat.so] エラー 1
rake aborted!
```

This is because `create_rust_makefile` always tries to pass ARGV to `cargo`. To be honest, I'm not sure why `create_rust_makefile` uses ARGV. But, I think `create_rust_makefile` only calls via other commands(e.g. Ruby, Rake).
So using ARGV only after `--` is commonplace I think.

### DIFF
--- a/gem/lib/rb_sys/mkmf.rb
+++ b/gem/lib/rb_sys/mkmf.rb
@@ -167,8 +167,8 @@ module RbSys
     def cargo_command(cargo_dir, builder)
       builder.ext_dir = cargo_dir
       dest_path = builder.target_dir || File.join(Dir.pwd, "target")
-      args = ARGV.dup
-      args.shift if args.first == "--"
+      args = []
+      args = ARGV.dup.shift if args.first == "--"
       cargo_cmd = builder.cargo_command(dest_path, args)
       cmd = Shellwords.join(cargo_cmd)
       cmd.gsub!("\\=", "=")


ianks:
Good catch, ty

Issue #424, opened 2024-09-23
Use stable_abi/compiled.{c,rs} on non-CRuby


eregon:
Hello,

I'm coming from https://github.com/oracle/truffleruby/issues/3396, in short we're trying to make Rust extensions work on TruffleRuby.
There were some minor fixes in RbConfig, etc which have been fixed in truffleruby-dev.

The remaining errors are in https://github.com/oracle/truffleruby/issues/3396#issuecomment-2356400143
So for example:
```
error[E0609]: no field `flags` on type `uncategorized::RBasic`
  --> /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/commonmarker-1.1.5/ext/commonmarker/.rb-sys/stable/cargo/registry/src/index.crates.io-6f17d22bba15001f/rb-sys-0.9.99/src/stable_api/ruby_3_2.rs:19:35
   |
19 |         let flags = rstring.basic.flags;
   |                                   ^^^^^ unknown field
```

This is because TruffleRuby [doesn't and cannot](https://github.com/oracle/truffleruby/issues/3118) expose `flags` on `struct RBasic`.
And anyway we can see the code in ruby_3_2.rs is highly CRuby-specific:
https://github.com/oxidize-rb/rb-sys/blob/991df6de21d244631e1b889070a8e4136f4220fe/crates/rb-sys/src/stable_api/ruby_3_2.rs#L19-L26

So such code in this file should only be used on CRuby, and on no other Ruby implementation.
And https://github.com/oxidize-rb/rb-sys/blob/991df6de21d244631e1b889070a8e4136f4220fe/crates/rb-sys/src/stable_api/compiled.c should be used instead.
A good way to test if it's CRuby is `RUBY_ENGINE == "ruby"`.

Could you fix this issue?

Related: https://github.com/oxidize-rb/rb-sys/pull/229 cc @ianks

ianks:
Hey @eregon so sorry I missed this until now. Working on making this happen -> https://github.com/oxidize-rb/rb-sys/pull/441

PR #441, opened 2024-11-17
Add truffleruby support

ianks:
Certain implementation details will differ on truffleruby, so we should always bind to the C API.

closes https://github.com/oxidize-rb/rb-sys/issues/424

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       fail-fast: false
       matrix:
         # Test against all versions supported by rubygems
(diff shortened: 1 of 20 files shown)


ianks:
Ok so I got basic gems building and the test suite passing on my mac. I had to skip the `cargo test` suite of tests since I got stuck on linking issues (missing symbols for things like `rb_cBasicObject` etc) but that's OK for now.

ianks:
released in [v0.9.103](https://github.com/oxidize-rb/rb-sys/releases/tag/v0.9.103)

PR #452, opened 2024-11-21
Add stable API for Ruby 3.4

ianks:
This PR adds a stable API implementation for Ruby 3.4. It is exactly the same as Ruby 3.3, since there were not incompatible changes in this release cycle for the methods we support.

### DIFF
--- a/crates/rb-sys-test-helpers/src/ruby_test_executor.rs
+++ b/crates/rb-sys-test-helpers/src/ruby_test_executor.rs
@@ -1,5 +1,6 @@
 use std::error::Error;
 use std::panic;
+use std::ptr::addr_of_mut;
 use std::sync::mpsc::{self, SyncSender};
 use std::sync::Once;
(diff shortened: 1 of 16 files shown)


crates/rb-sys-test-helpers/src/ruby_test_executor.rs
```diff
@@ -141,6 +142,9 @@ pub unsafe fn setup_ruby_unguarded() {
         rb_sys::rb_w32_sysinit(&mut argc, &mut argv);
     }
 
+    let mut stack_marker: VALUE = 0;
+    ruby_init_stack(addr_of_mut!(stack_marker) as *mut _);
```
ianks, line 146: This was causing GC issues in Ruby 3.4, due to Ruby being unable to mark objects on the machine stack. Not sure why this wasn't an issue before, but now it's fixed.

Issue #450, opened 2024-11-21
Implement new `StableApiDefinition::thread_sleep`


ianks:
In order to support Truffleruby on magnus, we need to support thread sleep operations, which currently relies on cruby internals.

```rust
/// Blocks the current thread until the given duration has passed.
fn thread_sleep(&self, duration: Duration);
```

PR #457, opened 2024-11-27
Implemented StableApiDefinition::thread_sleep for TruffleRuby.

goyox86:
Closes https://github.com/oxidize-rb/rb-sys/issues/450

Very naive implementation, I think we probably should be using something that allows us to go sub-second. someting that takes a pointer to `rb_cTime or directly a timeval.

### DIFF
--- a/crates/rb-sys-tests/src/stable_api_test.rs
+++ b/crates/rb-sys-tests/src/stable_api_test.rs
@@ -647,3 +647,11 @@ parity_test!(
     },
     expected: true
 );
+
+parity_test!(
+    name: test_rb_thread_sleep,
+    func: thread_sleep,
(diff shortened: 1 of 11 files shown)


ianks:
@goyox86 can we use [`rb_thread_wait_for`](https://github.com/ruby/ruby/blob/bf225feb265af6d796721d3a7e2b6800d160d42d/include/ruby/internal/intern/thread.h#L199)??

goyox86:
> @goyox86 can we use [`rb_thread_wait_for`](https://github.com/ruby/ruby/blob/bf225feb265af6d796721d3a7e2b6800d160d42d/include/ruby/internal/intern/thread.h#L199)??

Done!

Issue #446, opened 2024-11-21
Implement new `StableApiDefinition` RBasic methods


ianks:
In order to support Truffleruby on magnus, we need to support RBasic operations, which currently relies on cruby internals.

```rust
/// Get the class from a VALUE which contains an RBasic struct.
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying RBasic struct. The caller must ensure that the
/// `VALUE` is a valid pointer to a non-immediate object.
unsafe fn rbasic_class(&self, obj: VALUE) -> VALUE;

/// Checks if the given object is frozen.
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying RBasic struct. The caller must ensure that the
/// `VALUE` is a valid pointer to a non-immediate object.
unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool;
```

PR #458, opened 2024-11-28
Implementing StableApiDefinition RBasic methods for TruffleRuby.

goyox86:
Closes https://github.com/oxidize-rb/rb-sys/issues/446

Not sure if calling `rb_obj_frozen_p` is correct in the Rust side. 

Tried to do what MRI does but `ruby_fl_type` does not seem to be available in bindgen's output.

### DIFF
--- a/crates/rb-sys/src/stable_api.rs
+++ b/crates/rb-sys/src/stable_api.rs
@@ -54,6 +54,16 @@ pub trait StableApiDefinition {
     /// is valid.
     unsafe fn rarray_const_ptr(&self, obj: VALUE) -> *const VALUE;
 
+    /// Get the class from a VALUE which contains an RBasic struct.
+    ///
+    /// `VALUE` is a valid pointer to a non-immediate object.
+    unsafe fn rbasic_class(&self, obj: VALUE) -> VALUE;
+
+    /// Checks if the given object is frozen.
+    ///
+    /// `VALUE` is a valid pointer to a non-immediate object.
+    unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool;
+
     /// Tests if a bignum is positive.
     ///
     /// # Safety
--- a/crates/rb-sys/src/stable_api.rs
+++ b/crates/rb-sys/src/stable_api.rs
@@ -62,7 +62,7 @@ pub trait StableApiDefinition {
     /// Checks if the given object is frozen.
     ///
     /// `VALUE` is a valid pointer to a non-immediate object.
-    unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool;
+    unsafe fn frozen_p(&self, obj: VALUE) -> bool;
 
     /// Tests if a bignum is positive.
     ///
--- a/crates/rb-sys/src/stable_api.rs
+++ b/crates/rb-sys/src/stable_api.rs
@@ -12,7 +12,10 @@
 //!    changes in production.
 
 use crate::VALUE;
-use std::os::raw::{c_char, c_long};
+use std::{
+    os::raw::{c_char, c_long},
+    ptr::NonNull,
+};
 
 pub trait StableApiDefinition {
     const VERSION_MAJOR: u32;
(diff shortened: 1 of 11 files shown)


crates/rb-sys/src/stable_api/ruby_3_0.rs
```diff
@@ -96,6 +96,26 @@ impl StableApiDefinition for Definition {
...
+    #[inline]
+    unsafe fn rbasic_class(&self, obj: VALUE) -> VALUE {
+        let rbasic = obj as *const crate::RBasic;
+
+        (*rbasic).klass
+    }
+
+    #[inline]
+    unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool {
+        // FIXME: Why this does not work?
```
goyox86, line 108: @ianks Please let me know if the `crate::rb_obj_frozen_p` version works to remove these comments.

ianks: Let's re-implement them in the versioned APIs and do the bit-fiddling ourselves (to avoid dylib dispatch overhead). If that breaks things for truffle for now, that's OK. I want to re-think that definition once we see issues surface

crates/rb-sys/src/stable_api/ruby_2_6.rs
```diff
@@ -88,6 +88,26 @@ impl StableApiDefinition for Definition {
...
+    }
+
+    #[inline]
+    unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool {
+        // FIXME: Why this does not work?
+        // if self.special_const_p(obj) {
+        //     true
+        // } else {
+        //     let rbasic = obj as *const crate::RBasic;
+        //     ((*rbasic).flags & crate::ruby_value_type::FL_FREEZE as VALUE) != 0
```
ianks, line 105: isn't it:

```suggestion
        //     ((*rbasic).flags & crate::ruby_fl_type::FL_FREEZE as VALUE) != 0
```

crates/rb-sys/src/stable_api/ruby_2_6.rs
```diff
@@ -88,6 +88,26 @@ impl StableApiDefinition for Definition {
...
 
+    #[inline]
+    unsafe fn rbasic_class(&self, obj: VALUE) -> VALUE {
+        let rbasic = obj as *const crate::RBasic;
+
+        (*rbasic).klass
+    }
+
+    #[inline]
+    unsafe fn rbasic_frozen_p(&self, obj: VALUE) -> bool {
```
ianks, line 99: On consideration, I think we should rename this to `frozen_p` instead and have it Just Work for immediates as well. Thoughts?

crates/rb-sys/src/stable_api.rs
```diff
@@ -54,6 +54,22 @@ pub trait StableApiDefinition {
     /// is valid.
     unsafe fn rarray_const_ptr(&self, obj: VALUE) -> *const VALUE;
 
+    /// Get the class from a VALUE which contains an RBasic struct.
+    ///
+    /// # Safety
+    /// This function is unsafe because it dereferences a raw pointer to get
+    /// access to underlying RBasic struct. The caller must ensure that the
+    /// `VALUE` is a valid pointer to a non-immediate object.
+    unsafe fn rbasic_class(&self, obj: VALUE) -> VALUE;
```
ianks, line 63: I'm wondering if we should handle the null case more explicitly (hidden class), so make this:

```suggestion
    unsafe fn rbasic_class(&self, obj: VALUE) -> Option<NonNull<VALUE>>;
```

crates/rb-sys/src/stable_api/ruby_2_6.rs
goyox86: Yup yup, typo. ruby_fl_type is not available I had to whitelist it in the bindings configuration to make it work for the bignum_p PRs.

crates/rb-sys/src/stable_api.rs
goyox86: Worth trying, that could be a benefit of the stable API: to leverage Rust type system when it makes sense to do so. Will give it a shot!

goyox86:
@ianks Applied the changes (and rebased) I think we should be good now.

ianks (approved):

Issue #451, opened 2024-11-21
Implement new `StableApiDefinition` bignum methods


ianks:
In order to support Truffleruby on magnus, we need to support bignum operations, which currently relies on cruby internals.

```rust
/// Tests if a bignum is positive.
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying RBasic struct. The caller must ensure that the
/// `VALUE` is a valid pointer to a bignum.
unsafe fn bignum_positive_p(&self, obj: VALUE) -> bool;

/// Tests if a bignum is negative.
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying RBasic struct. The caller must ensure that the
/// `VALUE` is a valid pointer to a bignum.
unsafe fn bignum_negative_p(&self, obj: VALUE) -> bool;
```

PR #461, opened 2024-12-02
Implemented StableApiDefinition::[bignum_positive_p|bignum_negative_p].

goyox86:
Closes https://github.com/oxidize-rb/rb-sys/issues/451

For the Rust implementation I choose a similar approach than CRuby.

### DIFF
--- a/crates/rb-sys-build/src/bindings.rs
+++ b/crates/rb-sys-build/src/bindings.rs
@@ -68,9 +68,7 @@ pub fn generate(
     let bindings = if cfg!(feature = "bindgen-deprecated-types") {
         bindings
     } else {
-        bindings
-            .blocklist_item("^ruby_fl_type.*")
-            .blocklist_item("^_bindgen_ty_9.*")
+        bindings.blocklist_item("^_bindgen_ty_9.*")
     };
 
     let bindings = opaqueify_bindings(rbconfig, bindings, &mut wrapper_h);
--- a/crates/rb-sys/src/stable_api.rs
+++ b/crates/rb-sys/src/stable_api.rs
@@ -54,6 +54,22 @@ pub trait StableApiDefinition {
     /// is valid.
     unsafe fn rarray_const_ptr(&self, obj: VALUE) -> *const VALUE;
 
+    /// Tests if a bignum is positive.
+    ///
+    /// # Safety
+    /// This function is unsafe because it dereferences a raw pointer to get
+    /// access to underlying RBasic struct. The caller must ensure that the
+    /// `VALUE` is a valid pointer to a bignum.
+    unsafe fn bignum_positive_p(&self, obj: VALUE) -> bool;
+
+    /// Tests if a bignum is negative.
+    ///
+    /// # Safety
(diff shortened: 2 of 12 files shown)


crates/rb-sys/src/stable_api/ruby_2_7.rs
```diff
@@ -96,6 +96,16 @@ impl StableApiDefinition for Definition {
         is_immediate || !test
     }
 
+    #[inline]
+    unsafe fn bignum_positive_p(&self, obj: VALUE) -> bool {
+        crate::rb_big_sign(obj) == 1
```
ianks, line 101: let's re-implement the flag checks to avoid dylib dispatch (and so rustc can inline more). This will keep the performance on par with [the C macros](https://github.com/ruby/ruby/blob/bf225feb265af6d796721d3a7e2b6800d160d42d/internal/bignum.h#L171-L175)

crates/rb-sys/src/stable_api.rs
```diff
@@ -132,6 +132,22 @@ pub trait StableApiDefinition {
...
+    /// `VALUE` is a valid pointer to a bignum.
+    unsafe fn bignum_positive_p(&self, obj: VALUE) -> bool;
+
+    /// Tests if a bignum is negative.
+    ///
+    /// # Safety
+    /// This function is unsafe because it dereferences a raw pointer to get
+    /// access to underlying RBasic struct. The caller must ensure that the
+    /// `VALUE` is a valid pointer to a bignum.
+    unsafe fn bignum_negative_p(&self, obj: VALUE) -> bool;
```
ianks, line 150: Let's just have a default impl here and remove manual impls on each versioned API:
```suggestion
    unsafe fn bignum_negative_p(&self, obj: VALUE) -> bool {
      !self.bignum_positive_p(ob)
    }
```

crates/rb-sys/src/stable_api/ruby_2_7.rs
ianks: Forget about the truffle failures for now, we'll circle back on those and add a new definition file

goyox86: You mean like this? I had to whitelist `ruby_fl_types` in order to bindgen to generate. let me know if this is what you meant 😅

crates/rb-sys-build/src/bindings.rs
```diff
@@ -68,9 +68,7 @@ pub fn generate(
     let bindings = if cfg!(feature = "bindgen-deprecated-types") {
         bindings
     } else {
-        bindings
-            .blocklist_item("^ruby_fl_type.*")
-            .blocklist_item("^_bindgen_ty_9.*")
+        bindings.blocklist_item("^_bindgen_ty_9.*")
```
goyox86, line 71: Not sure how else would I have gotten ruby_fl_type in the bindings 😅

crates/rb-sys/src/stable_api/ruby_2_7.rs
ianks: yes exactly... suprised to see they were gone..... 🤔

PR #467, opened 2024-12-16
WIP: Implementing RStruct methods for `StableApiDefinition` in order to support TruffleRuby

goyox86:
Closes https://github.com/oxidize-rb/rb-sys/issues/448

Implementing RStruct methods for `StableApiDefinition` in order to support TruffleRuby.

### DIFF
--- a/crates/rb-sys-build/src/bindings/wrapper.h
+++ b/crates/rb-sys-build/src/bindings/wrapper.h
@@ -69,3 +69,19 @@
 #ifdef HAVE_RUBY_IO_BUFFER_H
 #include "ruby/io/buffer.h"
 #endif
+
+struct RStruct {
+  struct RBasic basic;
+  union {
+    struct {
+      long len;
+      const VALUE *ptr;
+    } heap;
+    /* This is a length 1 array because:
+     *   1. GCC has a bug that does not optimize C flexible array members
+     *      (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102452)
+     *   2. Zero length arrays are not supported by all compilers
+     */
+    const VALUE ary[1];
+  } as;
+};
--- a/crates/rb-sys/src/stable_api/ruby_3_3.rs
+++ b/crates/rb-sys/src/stable_api/ruby_3_3.rs
@@ -1,13 +1,24 @@
 use super::StableApiDefinition;
 use crate::{
     internal::{RArray, RString},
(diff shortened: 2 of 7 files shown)


crates/rb-sys/src/stable_api/ruby_3_3.rs
```diff
@@ -268,4 +279,37 @@ impl StableApiDefinition for Definition {
...
+    // }
+
+    unsafe fn rstruct_len(&self, st: VALUE) -> c_long {
+        let rbasic = st as *const crate::RBasic;
+        if ((*rbasic).flags & RSTRUCT_EMBED_LEN_MASK as VALUE) != 0 {
+            let mut ret = ((*rbasic).flags & RSTRUCT_EMBED_LEN_MASK as VALUE) as c_long;
+            ret >>= RSTRUCT_EMBED_LEN_SHIFT;
+            ret
+        } else {
+            let rstruct = st as *const crate::RStruct;
```
goyox86, line 311: Hey @ianks need you help here!

I implemented `rstruct_len` using  `rb_struct_size` and that works fine but then in order to keep perf inline with macros I implemented it by hand using CRuby's implementation but I am struggling to find a way of tweaking the bindings so I get `RStruct`.

Got any pointers?

I did declare `RStruct` in wrapper.h to make it work but that is not probably how it should be done.

Anyways, let me know!

ianks: You're doing it right. `RStruct` is hidden, so we'll have to manually "bindgen" it hard coding it's `#[repr(C)]` equivalent in rust, like you have already done.

crates/rb-sys-build/src/bindings/wrapper.h
```diff
@@ -69,3 +69,19 @@
 #ifdef HAVE_RUBY_IO_BUFFER_H
 #include "ruby/io/buffer.h"
 #endif
+
+struct RStruct {
```
ianks, line 73:
```suggestion
#[repr(C)]
struct RStruct {
```

crates/rb-sys/src/stable_api/ruby_3_3.rs
goyox86: Manually bindgen struct `rstruct_len` and `struct_get` seem to be working fine, replaced the dynamically link ones by handwritten ones. Two remaining.

PR #481, opened 2025-01-08
Remove example Cargo.lock from git

ianks:


ianks:
this is going to pass

PR #482, opened 2025-01-08
Update bindgen to 0.71

jbourassa:
Bump bindgen to 0.71 in hope to fix compilation on Windows with recent clang. IIUC, latest bindgen should not have that problem.

I ran into 2 issues when bumping the crate:
1. ABI checks that referenced either private fields or out-of-scope structs:
    ```rust
    const _: () = {
        ["Size of RString"][::std::mem::size_of::<RString>() - 40usize];
        ["Alignment of RString"][::std::mem::align_of::<RString>() - 8usize];
        ["Offset of field: RString::basic"][::std::mem::offset_of!(RString, basic) - 0usize];
        ["Offset of field: RString::len"][::std::mem::offset_of!(RString, len) - 16usize];
        ["Offset of field: RString::as_"][::std::mem::offset_of!(RString, as_) - 24usize];
    };
    ```
  
    I fixed it by patching those blocks with a `use crate::internal::*;` and getting rid of the `rb_sys__Opaque__` prefix. Not sure what underlying change made this a problem though.

2. Missing `proc_macro2::Literal::c_string` when testing against magnus. Hence the proc_macro2 bump.

With those changes, magnus tests pass on my machine. Let's see what CI says.

### DIFF
--- a/crates/rb-sys-build/src/bindings/stable_api.rs
+++ b/crates/rb-sys-build/src/bindings/stable_api.rs
@@ -75,6 +75,21 @@ pub fn categorize_bindings(syntax: &mut syn::File) {
                         }
                     };
                 }
+            } else if let syn::Item::Const(ref mut c) = item {
+                if c.ident == syn::Ident::new("_", c.ident.span()) {
+                    let body = &mut c.expr;
+                    let code = body.clone().to_token_stream().to_string();
+                    let new_code = code.replace("rb_sys__Opaque__", "");
+                    let new_code = syn::parse_str::<syn::Expr>(&new_code).unwrap();
+
+                    *body = syn::parse_quote! {
+                        {
+                            #[allow(unused_imports)]
+                            use crate::internal::*;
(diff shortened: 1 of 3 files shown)


crates/rb-sys-build/src/bindings/stable_api.rs
```diff
@@ -75,6 +75,21 @@ pub fn categorize_bindings(syntax: &mut syn::File) {
...
+                    let new_code = syn::parse_str::<syn::Expr>(&new_code).unwrap();
+
+                    *body = syn::parse_quote!{
+                        {
+                            #[allow(unused_imports)]
+                            use crate::internal::*;
+                            #new_code
+                        }
+                    };
+                }
```
jbourassa, line 92: Not familiar with bindgen nor syn -- hopefully this makes sense. I think it preserves the intent, that is, enusre the types are ABI-compatible.

Scutua (approved):

PR #489, opened 2025-01-14
Bump rake-compiler-dock images to v1.8.0 with Ruby patch versions

stanhu:
This reflects the changes made in https://github.com/rake-compiler/rake-compiler-dock/pull/135.

While this requires packages to update their `RUBY_CC_VERSION` to use Ruby patches across the board, this is already needed for Ruby 3.3.5. This change at least makes this consistent.

See https://github.com/rake-compiler/rake-compiler-dock/releases/tag/v1.8.0.

### DIFF
--- a/docker/Dockerfile.aarch64-linux
+++ b/docker/Dockerfile.aarch64-linux
@@ -1,6 +1,6 @@
-FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.7.0-mri-aarch64-linux
+FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.8.0-mri-aarch64-linux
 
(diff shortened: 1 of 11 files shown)


Scutua (approved):

ianks:
Thank you so much! Running a build to check that it works, then will merge ❤️

https://github.com/oxidize-rb/rb-sys/actions/runs/12814269578

ianks:
Ahh, forgot that we need to make rb-sys-dock aware of the patch versions. Can you add them [here](https://github.com/oxidize-rb/rb-sys/blob/e7681aed605747d3905aa096d5bd0cd9906d030a/gem/exe/rb-sys-dock#L20-L22)?

(also, it's a little janky since we have to define the same string everywhere, open to ideas for improving as well)

stanhu:
Yes, I'll add them now.

PR #491, opened 2025-01-16
Support aarch64-linux-gnu and x86_64-linux-gnu builds

stanhu:
rake-compiler-dock now makes a clear distinction between GNU and Musl builds:
https://github.com/rake-compiler/rake-compiler-dock?tab=readme-ov-file#linux-gnu-and-musl-important-details

This change makes it possible to build images using the explicit aarch64-linux-gnu and x86_64-linux-gnu targets needed by rake-compiler-dock.


stanhu:
@ianks Will this change create the `rbsys/aarch64-linux-gnu` Docker image?

```
🐳 Downloading container "rbsys/aarch64-linux-gnu:0.9.108", this might take awhile...
❌ Error running command: $ ["/usr/local/bin/docker pull rbsys/aarch64-linux-gnu:0.9.108 --platform linux/amd64 --quiet > /dev/null"]
Error response from daemon: pull access denied for rbsys/aarch64-linux-gnu, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

ianks:
By default... I don't think so? Queued up a build to check -> https://github.com/oxidize-rb/rb-sys/actions/runs/12814145166

We probably want to add support for alias docker tags for this, and just push multiple tags. Maybe something like:

```
    {
      "ruby-platform": "aarch64-linux",
      "aliases": ["aarch64-linux-gnu", "aarch64-unknown-linux-gnu"],
      "rust-target": "aarch64-unknown-linux-gnu",
      "dockerfile": "docker/Dockerfile.aarch64-linux",
      "rake-compiler-dock": {
        "cc": "aarch64-linux-gnu-gcc"
      },
      "docker-platform": "linux/arm64/v8",
      "supported": true
    },
```

Then adjust the github workflow to push all the tags... Thoughts?

stanhu:
That sounds like a good approach.

stanhu:
I'm going to close this as I think the alias approach is the way to go.

PR #496, opened 2025-01-20
Clean up Ruby patch release overrides

stanhu:
`RakeCompilerDock.cross_rubies` supplies the exact patch release needed. This is available in rack-compiler-dock v1.9.1:

* https://github.com/rake-compiler/rake-compiler-dock/issues/149
* https://github.com/rake-compiler/rake-compiler-dock/releases/tag/v1.9.1

This simplifies the `Dockerfile`, but it does mean we have to keep the rake-compiler-dock version in the `Gemfile` in sync with all the images in `docker/Dockerfile.*`.


docker/Dockerfile.aarch64-linux
```diff
@@ -1,6 +1,7 @@
-FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.8.0-mri-aarch64-linux
+# Ensure this version matches the rack-compiler-version in Gemfile
+FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.9.1-mri-aarch64-linux
```
stanhu, line 2: Should we template these files?

Gemfile
```diff
@@ -8,6 +8,7 @@ gemspec path: "examples/rust_reverse"
 gem "rake", "~> 13.0"
 gem "minitest", "5.15.0"
 gem "rake-compiler", "~> 1.2.5" # Small bug in 1.2.4 that breaks Ruby 2.5
+gem "rake-compiler-dock", "~> 1.9.1" # This should match the versions used in docker/Dockerfile.*
```
gjtorikian, line 11: Nit, but shouldn't this be

```suggestion
gem "rake-compiler-dock", "1.9.1" # This should match the versions used in docker/Dockerfile.*
```

that way the version is absolutely pinned to the Dockerfile.

stanhu: Yes, good point, thanks.

docker/Dockerfile.aarch64-linux
ianks: One downside is that dependabot has a limit of PRs it will open, and it can be hard to know if all dockerfiles were updated properly.

ianks (approved):

ianks:
✅  Ty ty! We can fix the static build in a follow up

stanhu:
As I explained in https://github.com/oxidize-rb/rb-sys/issues/497, I believe the static build will be fixed by https://github.com/oxidize-rb/actions/pull/52. I think we just need someone to release a new version?

PR #499, opened 2025-01-28
Add rake-compiler-dock as a dependency

stanhu:
`gem/exe/rb-sys-dock` needs rake-compiler-dock to load the right Ruby versions, so require it as a dependency.


ianks:
accidentally beat you to it 😅 https://github.com/oxidize-rb/rb-sys/commit/0891b175c7af322747a609658288df4263c89ec9

PR #501, opened 2025-01-28
Ensure RUBY_CC_VERSION is always passed when building oxi-test

stanhu:
Previously it was possible with rake-compiler-dock to omit `RUBY_CC_VERSION`, but now it appears that the versions must be specified to  work. We default to all supported versions.


ianks:
Great catch

Issue #510, opened 2025-01-30
RUBY_TARGET not set properly when using `rb-sys-dock --platform x86_64-linux-gnu`


stanhu:
In [this CI job](https://github.com/y-crdt/yrb/actions/runs/13046206597/job/36397116089?pr=221,), the `oxidize-rb/actions/cross-gem@v1` action runs:

```
 rb-sys-dock --platform x86_64-linux-gnu --directory . --ruby-versions 3.1,3.2,3.3,3.4 --build
``` 

This now correctly pulls down the image `rbsys/x86_64-linux-gnu:0.9.109`, which is an alias for `rbsys/x86_64-linux:0.9.109`.

However, the `RUBY_TARGET` is set here:

https://github.com/oxidize-rb/rb-sys/blob/c1675472947306769a217bc0c4b684a1601d2640/docker/Dockerfile.x86_64-linux#L4

This should actually be `x86_64-linux-gnu`, not `x86_64-linux`. As a result, the build ends in failure:

```
Don't know how to build task 'native:x86_64-linux' (See the list of available tasks with `rake --tasks`)
Did you mean?  native:x86_64-linux-gnu
/tmp/rb-sys-dock/bundle/ruby/3.4.0/gems/rake-13.2.1/exe/rake:27:in '<top (required)>'
/usr/local/rbenv/versions/3.4.1/bin/bundle:25:in 'Kernel#load'
/usr/local/rbenv/versions/3.4.1/bin/bundle:25:in '<main>'
``` 

This distinction is important because gems compiled with `x86_64-linux` can work both on GNU and musl systems, but `x86_64-linux-gnu` is intended for GNU systems. 

This probably can be solved by:

1. Changing `RUBY_TARGET` from an `ENV` to an `ARG`.
2. Having `rb-sys-dock` pass in `-e RUBY_TARGET`.

stanhu:
Actually, it may be simpler just to set the `RUBY_TARGET` to `x86_64-linux-gnu` since that is what is expected from the image.

stanhu:
It seems that `/etc/rubybashrc` sets `RUBY_TARGET`, so it's not sufficient just to override the value with `-e` at runtime: https://github.com/oxidize-rb/rb-sys/blob/23ebd1e5d95daa3afcee0ba6ae888e65111b597d/docker/setup/rubybashrc.sh#L55

PR #511, opened 2025-01-30
Append -gnu to RUBY_TARGET for GNU Linux builds

stanhu:
rake-compiler-dock images now make a distinction between GNU and musl builds. Add a `-gnu` extension to `RUBY_TARGET` to make it possible to run:

```
rb-sys-dock --platform x86_64-linux-gnu --directory . --build
```

Closes https://github.com/oxidize-rb/rb-sys/issues/510


ianks:
note to self: check build result -> https://github.com/oxidize-rb/rb-sys/actions/runs/13057340828

stanhu:
Oh, sorry, the tests aren't passing because the image is producing `oxi-test-0.1.0-x86_64-linux-gnu.gem`, but it's expecting `oxi-test-0.1.0-x86_64-linux.gem`.

Perhaps we have to make this configurable then.

PR #512, opened 2025-01-30
Make it possible to override RUBY_TARGET

stanhu:
This reverts #511 because the CI test expected that `rb-sys-dock --platform x86_64-linux --directory . --build` produces a gem with a suffix of `-x86_64-linux.gem`.

This change drops the hard-coded `RUBY_TARGET` from the bash script so the value can be derived from the `ENV` setting. With this change, this is now possible:

1. `rb-sys-dock --platform x86_64-linux --directory . --build` produces a gem with a filename `-x86_64-linux-gnu.gem`.
1. `rb-sys-dock --platform x86_64-linux-gnu --directory . --build` produces a gem with a filename `-x86_64-linux-gnu.gem`.

### DIFF
--- a/docker/Dockerfile.aarch64-linux
+++ b/docker/Dockerfile.aarch64-linux
@@ -1,7 +1,7 @@
 # Ensure this version matches the rack-compiler-version in Gemfile
 FROM ghcr.io/rake-compiler/rake-compiler-dock-image:1.9.1-mri-aarch64-linux
 
-ENV RUBY_TARGET="aarch64-linux-gnu" \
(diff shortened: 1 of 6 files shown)


ianks (approved):

stanhu:
I seem to be making things worse here: https://github.com/oxidize-rb/rb-sys/actions/runs/13059416829/job/36438458768

I'm not sure why `RUBY_TARGET` is duplicated in so many places. At this point it almost seems easier to create a new Dockerfile for the GNU builds and drop this alias business.

PR #513, opened 2025-01-30
Allow RUBY_TARGET to be inherited from sudo

stanhu:
Previously RUBY_TARGET was never passed along because `sudo -u` was run. Add this variable to the sudoers file to preserve this varaible.

### DIFF
--- a/docker/setup/rb-sys-dock.sh
+++ b/docker/setup/rb-sys-dock.sh
@@ -3,7 +3,7 @@
 set -e
 
 setup_sudoers() {
-  echo "Defaults        env_keep += \"BUNDLE_PATH RB_SYS_CARGO_TARGET_DIR RAKEOPT\"" >> /etc/sudoers.d/rb-sys-dock
+  echo "Defaults        env_keep += \"BUNDLE_PATH RB_SYS_CARGO_TARGET_DIR RAKEOPT RUBY_TARGET\"" >> /etc/sudoers.d/rb-sys-dock
 }
 
 main() {


stanhu:
For some reason, I'm seeing this when I test this locally:

```
$ bundle exec rb-sys-dock --platform x86_64-linux-gnu --tag test --build
🐳 Using version test of the Docker image
platform is x86_64-linux-gnu
🐳 Running default build command (rake native:x86_64-linux-gnu gem)
Bundle complete! 11 Gemfile dependencies, 23 gems now installed.
Bundled gems are installed into `/tmp/rb-sys-dock/bundle`
rake aborted!
Don't know how to build task 'native:x86_64-linux-gnu' (See the list of available tasks with `rake --tasks`)
/tmp/rb-sys-dock/bundle/ruby/3.4.0/gems/rake-13.2.1/exe/rake:27:in '<top (required)>'
/usr/local/rbenv/versions/3.4.1/bin/bundle:25:in 'Kernel#load'
/usr/local/rbenv/versions/3.4.1/bin/bundle:25:in '<main>'
(See full trace by running task with --trace)
rake aborted!
Don't know how to build task 'native:x86_64-linux-gnu' (See the list of available tasks with `rake --tasks`)

(See full trace by running task with --trace)
```

stanhu:
I guess that problem happens even on the tagged versions, so I might be doing something wrong.

gem/exe/rb-sys-dock
```diff
@@ -310,8 +314,8 @@ def default_command_to_run(input_args)
   input_cmd = input_args.empty? ? "true" : input_args.join(" ")
 
   if OPTIONS[:build]
-    with_bundle = +"test -f Gemfile && bundle install && #{input_cmd} && bundle exec rake native:$RUBY_TARGET gem"
-    without_bundle = "#{input_cmd} && rake native:$RUBY_TARGET gem"
+    with_bundle = +"test -f Gemfile && bundle install && #{input_cmd} && bundle exec rake native:#{ruby_target} gem RUBY_TARGET=#{ruby_target}"
```
stanhu, line 317: I'm not quite sure why adding the environment variable at the end works, but why adding it via `export RUBY_TARGET=#{ruby_target} &&` before this line does not.

stanhu: `NAME=VALUE` feature is a well-documented `rake` feature: 
1. https://github.com/ruby/rake/blob/09a1cd41d13d0c2c23a0775f860a8c5e71d43911/doc/command_line_usage.rdoc#label-Rake+Command+Line+Usage
2. https://github.com/ruby/rake/blob/master/doc/rakefile.rdoc#label-Tasks+with+Arguments

In another Rake task, I confirmed `NAME=VALUE bundle exec rake` doesn't work.

ianks:
We

> Oh, I see:
> 
> 1. The `runas` script quietly runs `/usr/bin/sudo -u "$USER" -H BASH_ENV=/etc/rubybashrc -- "$@"`. This clears all environment varaiables and uses the `rubybashrc` script to set them up.
> 2. `RUBY_TARGET` is needed inside `./rb-sys/gem/lib/rb_sys/extensiontask.rb` to determine the cross-compile target.
> 
> `RUBY_TARGET` has been a static value, but now we're trying to make it dynamic. This does make me think that it might be easier to just create a separate `Dockerfile` here.

We may need to add it to sudoers.d?

```
$ docker run --platform linux/amd64 --rm -it rbsys/aarch64-linux:0.9.109 /bin/bash
root@c4a3fea5b9b9:/# cat /etc/sudoers.d/rb-sys-dock        
Defaults        env_keep += "BUNDLE_PATH RB_SYS_CARGO_TARGET_DIR RAKEOPT"
```

ianks:
We should probably make it include the defaults from rake-compiler-dock too...

```
root@c4a3fea5b9b9:/# cat /etc/sudoers.d/rake-compiler-dock 
Defaults        env_keep += "http_proxy https_proxy ftp_proxy GEM_PRIVATE_KEY_PASSPHRASE RCD_HOST_RUBY_PLATFORM RCD_HOST_RUBY_VERSION RCD_IMAGE RUBY_CC_VERSION LD_LIBRARY_PATH DEVTOOLSET_ROOTPATH SOURCE_DATE_EPOCH"
```

stanhu:
Yeah, I solved it by directly inserting `RUBY_TARGET` into the `Rakefile`, which bypasses the need to set this all up in the environment.

ianks:
> Yeah, I solved it by directly inserting `RUBY_TARGET` into the `Rakefile`, which bypasses the need to set this all up in the environment.

The one downside is that things wont work properly without `--build`, although maybe not a big deal

stanhu:
Ah, interesting. Let me try this out.

stanhu:
Thanks for that tip. I think this works.

ianks (approved):

stanhu:
@ianks Thanks for your help! Could you tag a new release so I can try upgrading a gem again?

PR #514, opened 2025-01-30
Revert "Make it possible to override RUBY_TARGET (#512)"

stanhu:
As far as I can tell, `RUBY_TARGET` needs to be a static value right now because of the way the Docker images are set up. This change broke the Docker images, so let's revert.

This reverts commit 7b037556cd53d9fed57e5048a289009a13bdbd17.


stanhu:
Oops, the revert is not complete.

stanhu:
Closing in favor of #513.

Issue #447, opened 2024-11-21
Implement new StableApi GC methods


ianks:
In order to support Truffleruby on magnus, we need to support GC operations, which currently relies on cruby internals.

```rust
/// Informs that there are external memory usages, so the GC can run more
/// often than it otherwise would if it was unaware of such allocations.
fn gc_adjust_memory_usage(&self, incr: isize);

/// Informs the GC that `young` is a new reference to `old`, allowing the
/// the old object to participate in generational GC.
fn gc_writebarrier(&self, old: VALUE, young: VALUE);

/// Opts out of generational GC and write barrier protection.
fn gc_writebarrier_unprotect(&self, obj: VALUE);
```

ianks:
See also #665 for additional context on RB_OBJ_WRITE write barrier implementation from Ruby's gc.h header.

eregon:
Does `gc_writebarrier_unprotect()` then make the object "shady" i.e. no write-barrier protected?
I'm not quite sure what it does from that description above.

Issue #443, opened 2024-11-21
Implement new StableApi string coderange methods


ianks:
In order to support Truffleruby on magnus, we need to support string coderange operations, which currently relies on cruby internals.

```rust
/// Gets the inline coderange of an encodable object (akin to `RB_ENC_CODERANGE`).
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying flags of the [`RBasic`]. The caller must ensure that
/// the `VALUE` is a valid pointer to an RString, or any other encodable object.
unsafe fn coderange_inline_get(&self, obj: VALUE) -> Self::CoderangeType;

/// Set the inline coderange of an encodable object (akin to `RB_ENC_CODERANGE_SET`).
///
/// **Notes**:
/// - A UTF-8 string can have either a 7-bit ASCII coderange if the string
///   is ASCII only. Only use the `RUBY_ENC_CODERANGE_VALID` value if the
///   string has non-ASCII characters to avoid breaking assumptions made
///   by the Ruby interpreter.
/// - This function does not check if the coderange is valid. It simply
///   sets the coderange to the given value. This caller must ensure that
///   the coderange is valid.
/// - This function does not modify the Ruby string's encoding. It only
///   sets the coderange.
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying flags of the [`RBasic`]. The caller must ensure that
/// the `VALUE` is a valid pointer to an RString, or any other encodable object.
unsafe fn coderange_inline_set(&self, obj: VALUE, coderange: Self::CoderangeType);


/// Clear the inline coderange of an encodable object (akin to `RB_ENC_CODERANGE_CLEAR`).
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying flags of the [`RBasic`]. The caller must ensure that
/// the `VALUE` is a valid pointer to an RString, or any other encodable object.
unsafe fn coderange_inline_clear(&self, obj: VALUE);

/// Mix  two code  ranges  into one.   This  is handy  for  instance when  you 
/// concatenate two  strings into one (akin to `RB_ENC_CODERANGE_AND`).
///
/// # Safety
/// This function is unsafe because it dereferences a raw pointer to get
/// access to underlying flags of the [`RBasic`]. The caller must ensure that
/// the `VALUE` is a valid pointer to an RString, or any other encodable object.
[...]

PR #516, opened 2025-02-06
Implemented GC methods for `StableApiDefinition` for TruffleRuby support.

goyox86:
Solves https://github.com/oxidize-rb/rb-sys/issues/447

Implemented:

- `StableApiDefinition::gc_adjust_memory_usage`.
- `StableApiDefinition::gc_writebarrier`.
- `StableApiDefinition::gc_writebarrier_unprotect`.

Note:

I have to tweak the testing macros so it would allow multiple arguments for the data_builder, in the meanwhile I just expanded the macro and tweaked it manually.

### DIFF
--- a/crates/rb-sys-tests/src/stable_api_test.rs
+++ b/crates/rb-sys-tests/src/stable_api_test.rs
@@ -405,6 +405,39 @@ parity_test! (
     }
 );
 
+#[rb_sys_test_helpers::ruby_test]
+fn test_rb_gc_writebarrier() {
+    use rb_sys::stable_api;
+    let old = ruby_eval!("String.new") as VALUE;
+    let young = ruby_eval!("42") as VALUE;
+    assert_ne!(stable_api::get_default().version(), (0, 0));
+    #[allow(unused)]
+    let rust_result = unsafe { stable_api::get_default().gc_writebarrier(old, young) };
+    #[allow(unused_unsafe)]
+    let compiled_c_result = unsafe { stable_api::get_compiled().gc_writebarrier(old, young) };
+    assert_eq!(
+        compiled_c_result, rust_result,
+        "compiled_c was {:?}, rust was {:?}",
+        compiled_c_result, rust_result
+    );
+}
+
+#[rb_sys_test_helpers::ruby_test]
+fn test_rb_gc_writebarrier_unprotect() {
+    use rb_sys::stable_api;
+    let obj = ruby_eval!("String.new") as VALUE;
+    assert_ne!(stable_api::get_default().version(), (0, 0));
+    #[allow(unused)]
+    let rust_result = unsafe { stable_api::get_default().gc_writebarrier_unprotect(obj) };
+    #[allow(unused_unsafe)]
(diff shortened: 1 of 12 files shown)


goyox86:
Looking at Truffle's failure I think they don't have a C API stub for `impl_gc_writebarrier`, `impl_gc_writebarrier_unprotect`.

eregon:
Sorry for the late reply, @nirvdrum pinged me today about this.

`rb_gc_adjust_memory_usage()` is defined on TruffleRuby so that one is fine.

For rb_gc_writebarrier() and rb_gc_writebarrier_unprotect() they are documented in https://github.com/ruby/ruby/blob/17f6a689629ec3507a79fc156833f329dc641bed/include/ruby/internal/gc.h#L660-L675
Copying it here for convenience:
```c
/**
 * This  is  the  implementation  of  #RB_OBJ_WRITE().   People  don't  use  it
 * directly.
 *
 * @param[in]   old    An object that points to `young`.
 * @param[out]  young  An object that is referenced from `old`.
 */
void rb_gc_writebarrier(VALUE old, VALUE young);

/**
 * This is the  implementation of #RB_OBJ_WB_UNPROTECT().  People  don't use it
 * directly.
 *
 * @param[out] obj  An object that does not participate in WB.
 */
void rb_gc_writebarrier_unprotect(VALUE obj);
```
So those should not be called directly according to those docs, and instead RB_OBJ_WRITE/RB_OBJ_WB_UNPROTECT should be used.
RB_OBJ_WRITE is supported on TruffleRuby. Can rb-sys use/expose that instead?
If not, see https://github.com/oracle/truffleruby/blob/b3a2cec22635f55b74f350bc59a9fbcca0e4ac3e/lib/cext/include/ruby/internal/gc.h#L791-L805 i.e. TruffleRuby has an empty rb_obj_written() and so rb_gc_writebarrier() if it existed would be empty too.

RB_OBJ_WB_UNPROTECT is not yet supported on TruffleRuby, I believe because so far we have not found any extension using that (do you know any extension actually using it?).
FWIW the docs of it:
```c
/**
 * Asserts that the passed object is  not fenced by write barriers.  Objects of
 * such  property do  not contribute  to  generational GCs.   They are  scanned
 * always.
 *
 * @param[out]  x  An object that would not be protected by the barrier.
 */
#define RB_OBJ_WB_UNPROTECT(x) rb_obj_wb_unprotect(x, __FILE__, __LINE__)
```

In general TruffleRuby currently never moves any native memory, and does not use a generational GC for it, so currently has no write barriers for writes in C extensions.

To be future-proof though it would be best to test if these functions exist (e.g. `have_func('rb_gc_writebarrier')` in extconf.rb). And if they don't then assuming they are empty are OK for write-barrier-related functions.

eregon:
Also don't hesitate to ping me and/or @andrykonchin on TruffleRuby-related issues in rb-sys.
We are happy rb-sys has truffleruby in CI now and want to help to keep it there.

ankane:
I think this and #443 should unlock support for Magnus on TruffleRuby.

Outstanding errors: https://github.com/matsadler/magnus/issues/165#issuecomment-4309826942

PR #518, opened 2025-02-13
Add simple retry to rustup commands in `rb_sys/mkmf`

ianks:
Starting to see some sporadic network failures in rustup downloads. We should retry in these cases.


ianks:
🙄

PR #530, opened 2025-05-06
Add stable API support for `RTypedData`

ianks:
In Ruby 3.5, the struct definition of `RTypedData` was [changed to optimize its size](https://github.com/ruby/ruby/pull/13190). To support this, this PR adds `rtypeddata_p`, `rtypeddata_embedded_p`, `rtypeddata_type`, and `rtypeddata_get_data` in StableApiDefinition for Ruby 2.6–3.4.


goyox86 (commented):
Late to the party, retroactive 👍🏼 😅

Issue #534, opened 2025-05-08
Can no longer build rb-sys on Raspberry Pi 32-bit platform


stanhu:
We're phasing out support for Raspberry Pi 32-bit, but #530 doesn't compile on a Raspberry Pi 32-bit platform:

```
error[E0308]: mismatched types
-->
/run/registry/src/index.crates.io-1cd66030c949c28d/rb-sys-0.9.113/src/stable_api/ruby_3_3.rs:296:23
    |
296 |         (typed_flag & (crate::TYPED_DATA_EMBEDDED as u64)) != 0
|                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`,
found `u64`
error[E0277]: no implementation for `u32 & u64`
-->
/run/registry/src/index.crates.io-1cd66030c949c28d/rb-sys-0.9.113/src/stable_api/ruby_3_3.rs:296:21
    |
296 |         (typed_flag & (crate::TYPED_DATA_EMBEDDED as u64)) != 0
    |                     ^ no implementation for `u32 & u64`
    |
    = help: the trait `BitAnd<u64>` is not implemented for `u32`
    = help: the following other types implement trait `BitAnd<Rhs>`:
              <u32 as BitAnd>
              <u32 as BitAnd<&u32>>
              <&'a u32 as BitAnd<u32>>
              <&u32 as BitAnd<&u32>>
Some errors have detailed explanations: E0277, E0308.
For more information about an error, try `rustc --explain E0277`.
The following warnings were emitted during compilation:
```

stanhu:
It looks like this was fixed in v0.9.114.

PR #535, opened 2025-05-08
Fix compile errors with 32-bit platforms

stanhu:
AFter upgrading to rb-sys v0.9.113, we found that #530 doesn't compile on 32-bit platform because the typed_data is a VALUE type, but it was always casted to a 64-bit value.  Fix this by casting TYPED_DATA_EMBEDDED as a VALUE.

Closes https://github.com/oxidize-rb/rb-sys/issues/534


stanhu:
Ah, this was fixed in v0.9.114 with #532.

PR #638, opened 2025-11-06
Support target-prefixed install_name_tool for cross-compilation

xiaoxipang:
Improves `install_name_tool` detection to support cross-compilation with osxcross.

When cross-compiling for macOS, tools like osxcross prefix binaries with the target triple (e.g., `aarch64-apple-darwin-install_name_tool`). This change checks for both the native and target-prefixed versions of `install_name_tool`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ianks (approved):
ty!

PR #652, opened 2025-11-25
feat: add Ruby 4.0 stable API support

seuros:
Tested with 4.0.0-preview2

The Ruby API changed slightly, but was deprecated since ruby 3.1.

We could back port the change if needed.


seuros:
@ianks 👍🏼

seuros:
@ianks anything else ?

ianks:
tysm!

PR #654, opened 2025-12-02
fix: Ignored `target` defined via cargo config file

### LINKED ISSUE #568: Support `target/<triple>/release` in the generated Makefile
If I specify `target` explicitly in my `~/.cargo/config.toml`:

```toml
[build]
target = "aarch64-apple-darwin"
```

then when installing a gem with Rust extension (`css_inline` in my case), the generated `Makefile` will point to `target/release/<name>.dylib` (on macOS in my case), however, the actual artifact is `target/aarch64-apple-darwin/release/<name>.dylib`. This leads to:

```
cp: target/release/libcss_inline.dylib: No such file or directory
```

Stranger6667:
Fixes #568 

Besides these tests, I've checked it with the `rust_reverse` example by adding `examples/rust_reverse/ext/rust_reverse/.cargo/config.toml`:

```
[build]
target = "x86_64-unknown-linux-gnu"
```

and building it.

I am very sorry for parsing toml with regexes :( Let me know if it would be better to pull the proper parser as a dependency (as per my understanding, no external dependencies is one of the goals?) - I also saw that the package name is parsed with a regex already (as a fallback). As of now, `cargo config` is unstable :(

P.S. I think that if a benchmark is defined first in `Cargo.toml`, then the name inference would take that instead of the package name

### DIFF
--- a/gem/lib/rb_sys/cargo/config.rb
+++ b/gem/lib/rb_sys/cargo/config.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module RbSys
+  module Cargo
+    # Reads Cargo configuration files to extract build settings.
+    module Config
+      class << self
+        # Returns the build target from Cargo configuration files.
+        # Searches from start_dir up to root, then checks CARGO_HOME.
+        #
+        # @param start_dir [String] the directory to start searching from
+        # @return [String, nil] the target triple, or nil if not configured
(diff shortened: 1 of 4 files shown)


seuros (approved):

ianks:
Thanks for the fix and the thorough tests, @Stranger6667 — much appreciated! 🙏

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #694, opened 2026-01-02
Add automated RubyGems/crates.io publishing

gjtorikian:
This PR is an attempt to make publishing to RubyGems and crates.io simpler and more secure. 

Ultiamtely, both  [RubyGems](https://guides.rubygems.org/trusted-publishing/) and  [crates.io](https://crates.io/docs/trusted-publishing) support Trusted Publishing. The idea is that rather than one individual using their own auth token to publish updates, the GitHub Actions workflow can request a short lived token from the provider, and use that to issue updates. This should make it simpler for newly merged PRs to be released across RubyGems, crates.io, and Dockerhub.

In the new release.yml here, I propose:

- a workflow dispatch with two inputs, `prepare`, and `release`. 
- once a new version is ready to be released, one would go to the GitHub Actions UI and toggle the workflow with `prepare: true`
- this examines the new version number defined in `gem/lib/rb_sys/version.rb`, generates a CHANGELOG entry, and opens a PR 
- the PR can be modified and merged at will

Afterwards:

- the same workflow is run with `release: true`
- this creates a new tag, pushes it, and publishes a GitHub release with that tag
- docker.yml recognizes the new tag, and pushes the images to Dockerhub
- as well, the `publish_crates` job triggers after the new release is made. this pushes `rb-sys-build`, waits for it to be indexed, then pushes `rb-sys`, to crates.io using OIDC
- once the crates are pushed, `publish_gem` publishes the gem using OIDC

There needs to be some configuration on the RubyGems and crates.io UI settings to "recognize" this workflow as the source of truth, but that ought to be it. This. brings along an audit trail for releases (via the GitHub Actions logs), credential rotation via OIDC, and supply chain security.


ianks:
Cool with this in theory, i assume i need to input credentials somewhere?

gjtorikian:
bumping this!

gjtorikian:
@ianks bumping this since I see you working 😆 

basically, the gem release process is tied to you, and only you

gjtorikian:
@ianks any thoughts here on automating yourself out of releases? I've also been using [release-please](https://www.dxheroes.io/insights/release-and-publish-a-new-ruby-sdk-version-automatically-a-step-by-step-guide) at work which is a less ad hoc system than this one -- idea is the same though.

PR #703, opened 2026-01-08
Test with ruby head

nobu:


ianks:
FWIW, we can remove the legacy magnus 0.5 tests

PR #724, opened 2026-04-11
Correct optional arguments to nilable arguments

bangseongbeom:
It seems that `Option<T>` does not allow omitting arguments. It's still possible to pass `nil` in Ruby.

Later, it would be nice to include an explanation of [scan_args](https://docs.rs/magnus/latest/magnus/scan_args/fn.scan_args.html)￼.


gjtorikian:
thank you!

PR #725, opened 2026-04-11
Fix mobile sidebar not showing up

bangseongbeom:
Fix an issue where the mobile sidebar did not open when the button was pressed on mobile.

For `mobileSidebar.toggle()` to work properly, replace `nav` with [`NavbarLayout`](https://github.com/facebook/docusaurus/blob/v3.9.2/packages/docusaurus-theme-classic/src/theme/Navbar/Layout/index.tsx).

It also removes the button's border and background using `clean-btn`.


gjtorikian:
thank you!

PR #726, opened 2026-04-11
Fix favicon.ico

bangseongbeom:
The current favicon.ico is the default Docusaurus icon, so I generated it based on favicon.svg.


gjtorikian:
great, thanks!

PR #727, opened 2026-04-13
Remove `json` from runtime dependencies

hsbt:
The json gem is bundled as a default gem since Ruby 1.9, so there is no need to declare it as an explicit runtime dependency.

Removing it avoids unnecessary native compilation of the `json` gem on every `rb_sys` install.

### DIFF
--- a/gem/rb_sys.gemspec
+++ b/gem/rb_sys.gemspec
@@ -27,8 +27,6 @@ Gem::Specification.new do |spec|
     spec.signing_key = File.expand_path("~/.ssh/gem-private_key.pem") if $0.end_with?("gem")
   end
 
-  spec.add_dependency("json", ">= 2")
-
   # Needed for `RakeCompilerDock.cross_rubies`
   spec.add_dependency("rake-compiler-dock", "1.11.0")
 


gjtorikian:
Not trying to be difficult here but will that always be the case? Many bundled gems (`base64`, `csv`, etc) were removed from stdlib. It seems to me the actual problem is the `json` gem not being distributed precompiled.

ianks:
> Not trying to be difficult here but will that always be the case? Many bundled gems (base64, csv, etc) were removed from stdlib. It seems to me the actual problem is the json gem not being distributed precompiled.

same question @hsbt

hsbt:
Thanks to merge this 🙏 

I can't promise anything about the future, but my view is that `json`/`psych` should only be moved to bundled gems if precompiled binaries are reliably provided.

That said, I don't think "json isn't distributed precompiled" should be framed as *the* problem. Maintaining precompiled binaries is a significant ongoing burden on gem maintainers, and it's not something to casually point to as the fix.

gjtorikian:
As a maintainer of precompiled gems, I agree! It requires careful work and consideration to get it right. 

But I was responding to your comment

> Removing it avoids unnecessary native compilation

This just feels like solving the problem for one use case when there’s a deeper (harder) problem to solve for.

PR #743, opened 2026-06-17
Fix target-dir rewrite for dashed paths

michaelherold:
The generated Makefile rewrites Cargo's concrete `--target-dir` path to `$(RB_SYS_CARGO_TARGET_DIR)`. The previous regex stopped when it saw any `--` in the already shell-joined path, which left the rest of the path as a bogus Cargo option.

This caused directories like `~/trees/michaelherold--fix-bug-in-foo/` to fail to compile as it would split the double-dash from the path, leaving a non-existent directory.

This change matches the target directory as a shell word instead, so paths containing `--` or spaces are fully replaced.


gjtorikian:
@ianks does MSRV/Cargo need to be changed and/or older versions dropped to get CI green?

Regex looks good, just nervous to merge without knowing what's going on with the build.

ianks:
@gjtorikian hmm maybe, or maybe re-gen cargo.lock with older version?

its probably ok to bump msrv though. general rule of thumb is to support whatever stable debian packages for rustc

